PR #5, opened 2018-10-04
testscript: make it compile under Go 1.10

rogpeppe:
Making it compile under earlier versions than 1.10 is a bit
harder, so punt for now.

### DIFF
--- a/testscript/exe.go
+++ b/testscript/exe.go
@@ -190,9 +190,14 @@ func (nopTestDeps) WriteProfileTo(name string, w io.Writer, debug int) error {
 func (nopTestDeps) ImportPath() string {
 	return ""
 }
-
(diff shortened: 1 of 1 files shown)


testscript/exe.go
```diff
@@ -190,9 +190,14 @@ func (nopTestDeps) WriteProfileTo(name string, w io.Writer, debug int) error {
...
 	return ""
 }
-
 func (nopTestDeps) StartTestLog(w io.Writer) {}
 
 func (nopTestDeps) StopTestLog() error {
 	return nil
 }
+
+// Note: WriteHeapProfile is not needed for Go 1.10.
```
mvdan: shouldn't this rather be "needed to build on Go 1.10, but not actually useful"?

rogpeppe: yeah, I had the comment wrong. fixed.

PR #12, opened 2018-10-23
ci: fix travis to force GO111MODULE=on

myitcv:


.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: go
+env:
+  - GO111MODULE=on
 go_import_path: github.com/rogpeppe/go-internal
-go: 
-  - "1.11"
+install: echo "Using Go modules; no install step required"
```
rogpeppe: no need for the quotes here; they're only confusing matters, I think.
Because I never trust YAML, I'd probably do:

    install: 'echo no install step required'

PR #13, opened 2018-10-23
ci: add step to ensure our modules files are tidy

myitcv:


.travis.yml
```diff
@@ -5,4 +5,8 @@ go_import_path: github.com/rogpeppe/go-internal
 install: "echo no install step required"
 go:
   - "1.11.x"
-script: go test ./...
+script:
+  - go test ./...
+  - go mod tidy
+  - go list all > /dev/null # https://github.com/golang/go/issues/27868#issuecomment-431413621
+  - bash -c "test -z \"$(git status --porcelain)\" || (git status; git diff && false)"
```
mvdan: Isn't there a simpler way to check that there aren't any changes? For example, I imagine `git diff --exit-code` would do the trick.

I also believe Travis uses bash as the default shell, so `bash -c` should be unnecessary.

myitcv: > Isn't there a simpler way to check that there aren't any changes? For example, I imagine git diff --exit-code would do the trick.

Does this work where a new file happens to be created as part of CI, and hence is untracked?

> I also believe Travis uses bash as the default shell, so bash -c should be unnecessary.

Will give that a try :)

mvdan: > Does this work where a new file happens to be created as part of CI, and hence is untracked?

What file could be created? If you rely on `go.mod` and `go.sum` being there for newly set up repos, it doesn't matter.

myitcv: That's just it, I don't know what file(s) might be created. But if any are created, and the tests only pass because those files were created in the course of being run, then that's what I'm trying to catch here.

Is that too defensive?

myitcv: I'll go with this (i.e. the status check) for now and we can review in a later PR.

PR #16, opened 2018-11-05
testscript: fix "hello world" doc example

mvdan:
Two simple issues kept this script from actually working. Update it via
scripts/hello.txt.

### DIFF
--- a/testscript/doc.go
+++ b/testscript/doc.go
@@ -46,11 +46,11 @@ create in the script's temporary file system before it starts executing.
 As an example:
 
 	# hello world
-	exec cat hello
-	stdout "hello world\n"
+	exec cat hello.text
+	stdout 'hello world\n'
 	! stderr .
 
-	-- hello.txt--
+	-- hello.text --
 	hello world
 
 Each script runs in a fresh temporary work directory tree, available to scripts as $WORK.


testscript/doc.go
```diff
@@ -46,11 +46,11 @@ create in the script's temporary file system before it starts executing.
 As an example:
 
 	# hello world
-	exec cat hello
-	stdout "hello world\n"
+	exec cat hello.text
```
myitcv: Minor preference for `.txt` ...

mvdan: I just copied what's in testdata. The point of not using `.txt` is that the test would overwrite itself :)

I slightly prefer this over a filel called just `hello`, because that's less obviously a file.

myitcv: > The point of not using .txt is that the test would overwrite itself :)

No, because this is all run within a temp directory `$WORK`. So the txtar is expanded in `$WORK` and your initial working directory is `$WORK`

Agreed having a suffix is better than just `hello`.

mvdan: Gotcha. Even then, I'd say it could be confusing for a hello world test to use a temporary file that happens to be named just like the test script iself.

Ultimately this is up to @rogpeppe - but whatever we decide, the two scripts should be identical.

rogpeppe: How about just using "hello" as a file name?

mvdan: Three people, three ways of naming one silly file :)

Roger, did you see our previous comments about no filename extension? If you still prefer `hello`, I'll just change it to that.

PR #18, opened 2018-11-18
testscript: support background processes

myitcv:
Taken from https://go-review.googlesource.com/c/go/+/141218


myitcv:
Created incorrectly. Will recreate.

PR #22, opened 2018-11-19
testscript: add cmpenv command to testing script language

myitcv:
Originally added in https://go-review.googlesource.com/c/147277


myitcv:
Merging per #20

PR #23, opened 2018-11-19
testscript: add goversion environment variable to testing script language

myitcv:
Originally added in https://go-review.googlesource.com/c/147280


myitcv:
Merging per #21

PR #25, opened 2018-11-25
testscript: add *TestScript.Setenv method

myitcv:
Allows environment variables to be set by Run-Param commands.

### DIFF
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -495,6 +495,17 @@ func (ts *TestScript) MkAbs(file string) string {
 	return filepath.Join(ts.cd, file)
 }
 
+// Setenv sets the value of the environment variable named by the key.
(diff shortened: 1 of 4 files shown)


testscript/testscript.go
```diff
@@ -495,6 +495,12 @@ func (ts *TestScript) MkAbs(file string) string {
 	return filepath.Join(ts.cd, file)
 }
 
+// Setenv sets the value of the environment variable named by the key.
+func (ts *TestScript) Setenv(key, value string) {
```
rogpeppe: Perhaps cmdEnv should also call this function instead of duplicating the logic?
Also, given that we're providing some access to the environment, should we consider providing the capability to get environment variables too?

myitcv: Good call on both fronts. Done

PR #30, opened 2018-11-28
ci: test on multiple OS-es

myitcv:


.gitattributes
```diff
@@ -0,0 +1,9 @@
+# By default, do not attempt to do any end-of-line conversion upon
+# checkin or checkout.
+*       -text
+
+# For specific file-types, force conversion to LF line endings. This
+# can be overridden below or in a more specific .gitattributes if, for
+# example, we want to allow a .txt file to contain a CRLF line ending
+*.go    text eol=lf
+*.txt   text eol=lf
```
mvdan: A perhaps simpler option would be just treating all files as non-text, like what Go does. Then you rely on any developers on Windows to have their editors set up to use LF and not CRLF. But I see the appeal of your setup - you enforce Go and text files to never use CRLF.

.travis.yml
```diff
@@ -1,12 +1,30 @@
...
 script:
   - go test ./...
   - go mod tidy
-  - go list all > /dev/null # https://github.com/golang/go/issues/27868#issuecomment-431413621
+
+  # https://github.com/golang/go/issues/27868#issuecomment-431413621
+  - go list all > /dev/null
+
+  - test -z "$(gofmt -d .)" || (gofmt -d . && false)
   - test -z "$(git status --porcelain)" || (git status; git diff && false)
```
mvdan: tip - all beyond "go test ./..." doesn't need to be run on all platforms. In particular, Mac and Windows are way slower than a Linux container, so I tend to run all extra checks (gofmt, go mod, etc) on Linux only. For example: https://github.com/mvdan/sh/blob/master/.travis.yml#L11

cmd/txtar-savedir/testdata/txtar-savedir-self.txt
```diff
@@ -4,7 +4,7 @@
 # stdout check
 txtar-savedir blah
 ! stderr .+
-stdout '^module example.com/blah$'
+stdout '^module example.com[/\\]blah$'
```
mvdan: Can module paths (import paths) contain backslashes?

internal/textutil/diff.go
```diff
@@ -0,0 +1,74 @@
+package textutil
```
mvdan: I presume that this isn't from the Go repo? I think we should start labelling what packages mirror the Go repo, and what packages are ours.

internal/textutil/diff.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	lines2 := strings.Split(text2, "\n")
+	lines2 = lines2[:len(lines2)-1] // remove empty string after final line
+
+	// Naive dynamic programming algorithm for edit distance.
+	// https://en.wikipedia.org/wiki/Wagner–Fischer_algorithm
+	// dist[i][j] = edit distance between lines1[:len(lines1)-i] and lines2[:len(lines2)-j]
+	// (The reversed indices make following the minimum cost path
+	// visit lines in the same order as in the text.)
+	dist := make([][]int, len(lines1)+1)
```
mvdan: FYI, there are ways to implement this in non-quadratic memory. Shouldn't matter if the inputs will always be relatively small, though.

.gitattributes
myitcv: Per the default at the top of this file, I agree that's exactly what we want.

But per https://github.com/golang/go/blob/2b58ca6e3de3d93817a4e6cc55378015eb3c2040/.gitattributes#L5-L6 we would then require that we have some additional check to verify that for `.go` and `.txt` files (there could be others that need to fall into this category) don't accidentally end up with CRLF.

So I think this is what we (for now!)

.travis.yml
myitcv: Thanks - this is good advice. Will make this change.

cmd/txtar-savedir/testdata/txtar-savedir-self.txt
myitcv: Thanks, I suspect this was the result of an overly aggressive `sed` earlier in refactoring.

internal/textutil/diff.go
myitcv: This is literally a copy-paste of Russ' implementation. I don't think, per his comment, he was looking to win any awards with his initial implementation.

mvdan: Gotcha :) this was just a drive-by comment, in case you had written this yourself. Hence my previous comment about whether this package was a copy-paste.

.travis.yml
myitcv: Actually, I'm going to hold off this for now; would seem to make sense to see the line endings fiasco settle a bit first. I'm still scarred from those experiences!

internal/textutil/diff.go
myitcv: No, it's not. And good point. I'll add a quick package comment along with a note on how it came to life.

myitcv:
@mvdan feedback addressed, PTAL.

txtar/archive_test.go
```diff
@@ -39,6 +39,16 @@ hello world`,
 			},
 		},
 	},
+	{
+		name: "basic",
```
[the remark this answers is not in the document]
myitcv: Done

testscript/scripts/command.txt
```diff
@@ -1,2 +0,0 @@
-printargs a b 'c d'
```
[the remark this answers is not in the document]
myitcv: Not sure... I've added back the test and the associated command

PR #33, opened 2018-12-12
cmd/txtar-x: new command

rogpeppe:
This allows the easy extraction of files from a txtar file on the command line.

Also add functionality to testscript to make it possible to set up the standard
input of an execed command.

### DIFF
--- a/cmd/txtar-x/extract.go
+++ b/cmd/txtar-x/extract.go
@@ -1,0 +1,108 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The txtar-x command extracts a txtar archive to a filesystem.
+//
+// Usage:
+//
+//	txtar-x [-C root-dir] saved.txt
+//
+// See https://godoc.org/github.com/rogpeppe/go-internal/txtar for details of the format
+// and how to parse a txtar file.
+//
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/rogpeppe/go-internal/txtar"
+)
+
+var (
+	extractDir = flag.String("C", ".", "directory to extract files into")
+)
+
+func usage() {
+	fmt.Fprintf(os.Stderr, "usage: txtar-x [flags] [file]\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	os.Exit(main1())
+}
+
+func main1() int {
+	flag.Usage = usage
+	flag.Parse()
+	if flag.NArg() > 1 {
+		usage()
+		return 2
+	}
+	log.SetPrefix("txtar-x: ")
+	log.SetFlags(0)
+
+	var a *txtar.Archive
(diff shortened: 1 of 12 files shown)


cmd/txtar-x/extract.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The txtar-extract command extracts a txtar archive to a filesystem.
```
myitcv: `txtar-x`, no?

cmd/txtar-x/extract.go
```diff
@@ -0,0 +1,108 @@
...
+//
+// Usage:
+//
+//	txtar-extract [-d root-dir] saved.txt
```
myitcv: Ditto.

myitcv: Also the commentary here on the flag is out of date.

cmd/txtar-x/extract.go
```diff
@@ -0,0 +1,108 @@
...
+func extractFile(f txtar.File) error {
+	path := filepath.Clean(filepath.FromSlash(f.Name))
+	if isAbs(path) || strings.HasPrefix(path, ".."+string(filepath.Separator)) {
+		return fmt.Errorf("outside parent directory")
+	}
+	path = filepath.Join(*extractDir, path)
+	if err := os.MkdirAll(filepath.Dir(path), 0777); err != nil {
+		return err
+	}
+	// Avoid overwriting existing files by using O_EXCL.
```
myitcv: If we are looking to emulate `tar`, why not overwrite the file?

cmd/txtar-x/extract_test.go
```diff
@@ -0,0 +1,40 @@
...
+func TestScripts(t *testing.T) {
+	testscript.Run(t, testscript.Params{
+		Dir: "testdata",
+		Cmds: map[string]func(ts *testscript.TestScript, neg bool, args []string){
+			"unquote": unquote,
+		},
+	})
+}
+
+func unquote(ts *testscript.TestScript, neg bool, args []string) {
```
myitcv: Nice approach. An alternative would be extend the `txtar` filename format in some way to indicate that the contents should be unquoted, e.g. 

```
-- > file.txtar --
>some comment
>
>-- foo --
>foo
>-- a/b/bar --
>bar
```

testscript/doc.go
```diff
@@ -150,6 +150,9 @@ The predefined commands are:
   test. At the end of the test, any remaining background processes are
   terminated using os.Interrupt (if supported) or os.Kill.
 
+  The standard input can be provided using the stdin command; this will be
```
myitcv: Superfluous "The" at the start of the sentence?

cmd/txtar-x/extract_test.go
rogpeppe: meh. I like the simplicity of the current syntax (no errors possible), and if it can be worked around this easily, then I don't there's a particular need to change it.

testscript/doc.go
rogpeppe: done

PR #38, opened 2018-12-19
lockedfile: initial commit

### LINKED ISSUE #35: lockedfile: add package
Basically, port https://go-review.googlesource.com/c/go/+/145178

myitcv:
Based on https://go-review.googlesource.com/c/go/+/145178

Taken from the Go repo as of f880efcc1666e2b99fbda644eeff258e4a15dd38

We also include a snap of internal/syscall/windows for the Windows-based
implementation.

Closes #35.


internal/syscall/windows/mksyscall.go
```diff
@@ -0,0 +1,7 @@
+// Copyright 2016 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package windows
+
+//go:generate go run $GOROOT/src/syscall/mksyscall_windows.go -output zsyscall_windows.go syscall_windows.go security_windows.go psapi_windows.go symlink_windows.go
```
rogpeppe: shouldn't this file be guarded by a windows build tag, so we don't try to run this if we run `go generate` on a non-windows box?

myitcv: Possibly, but on the basis all I did was copy paste... I'm not sure

https://github.com/golang/go/blob/a2bb68de4d1cb1ca35279523365e9caf36fb9896/src/internal/syscall/windows/mksyscall.go

PR #40, opened 2019-01-04
gotooltest: add support for version constraints

myitcv:
This effectively mirrors the go/build constraint supports such as go1.12
to check that the Go version is >= 1.12.

Also remove some duplicated code in testscript that I had previously
missed was already present in gotooltest.

### DIFF
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -367,8 +367,6 @@ Script:
 // condition reports whether the given condition is satisfied.
 func (ts *TestScript) condition(cond string) (bool, error) {
 	switch cond {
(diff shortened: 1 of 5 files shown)


testscript/testscript.go
```diff
@@ -367,8 +367,6 @@ Script:
 // condition reports whether the given condition is satisfied.
 func (ts *TestScript) condition(cond string) (bool, error) {
 	switch cond {
-	case runtime.GOOS, runtime.GOARCH, runtime.Compiler:
```
myitcv: This is already present in gotooltest

PR #41, opened 2019-01-07
cmd/testscript: initial commit

myitcv:


cmd/testscript/main.go
```diff
@@ -0,0 +1,249 @@
...
+	// running a single script in a testscript instance, which means that we
+	// will only have a single subtest.
+	f(r)
+}
+
+func (r runner) Verbose() bool {
+	return r.verbose
+}
+
+func run(rd, fn string) error {
```
rogpeppe: I had to think a bit about what these names might be. Perhaps s/rd/runDir/ ?
Or a doc comment?

cmd/testscript/main.go
```diff
@@ -0,0 +1,249 @@
...
+	var err error
+
+	testdata := filepath.Join(rd, "testdata")
+	mods := filepath.Join(testdata, "goproxytest")
+
+	if err := os.MkdirAll(mods, 0777); err != nil {
+		return fmt.Errorf("failed to create testdata dir: %v", err)
+	}
+
+	if fn == os.Stdin.Name() {
```
rogpeppe: I'm not sure about this. On linux this will happen automatically (opening `/dev/stdin` reads stdin). On windows, who knows what os.Stdin.Name() returns? I'd either go with a standard name (`-`) or just say "if there are no files on the command line, read from stdin".

cmd/testscript/main.go
```diff
@@ -0,0 +1,249 @@
...
+
+	var toWrite []txtar.File
+
+	for _, f := range ar.Files {
+		fp := filepath.Clean(filepath.FromSlash(f.Name))
+		if isAbs(fp) || strings.HasPrefix(fp, ".."+string(filepath.Separator)) {
+			return fmt.Errorf("file %v would result in file outside parent directory", f.Name)
+		}
+
+		// only write out goproxytest rooted files
```
rogpeppe: // Save non-goproxytest files for later.

?

cmd/testscript/main.go
```diff
@@ -0,0 +1,249 @@
...
+		if err := os.MkdirAll(filepath.Dir(fp), 0777); err != nil {
+			return err
+		}
+		// Avoid overwriting existing files by using O_EXCL.
+		out, err := os.OpenFile(fp, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0666)
+		if err != nil {
+			return err
+		}
+
+		// unquote, because this will be a quoted goproxytest file
```
rogpeppe: I don't think this is necessary any more.

cmd/testscript/main.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to txtar parse %v: %v", fn, err)
+	}
+
+	var toWrite []txtar.File
+
+	for _, f := range ar.Files {
+		fp := filepath.Clean(filepath.FromSlash(f.Name))
+		if isAbs(fp) || strings.HasPrefix(fp, ".."+string(filepath.Separator)) {
```
rogpeppe: As it's a bit subtle and we don't really need any custom extraction logic, I wonder if this logic could be moved into go-internal/txtar now rather than duplicating it from cmd/txtar-x.

cmd/testscript/main.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+	}
+
+	ar.Files = toWrite
+
+	if err := ioutil.WriteFile(filepath.Join(testdata, "script.txt"), txtar.Format(ar), 0666); err != nil {
+		return fmt.Errorf("failed to write script for %v: %v", fn, err)
+	}
+
+	srv, err := goproxytest.NewServer(mods, "")
```
rogpeppe: This should be shut down after the test has run, I think.

Also, perhaps don't start the proxy when there are no goproxytest files?

cmd/testscript/README.md
```diff
@@ -0,0 +1,111 @@
...
+reproductions of command sequences. It is, in many respects, like a poor man's
+Docker.
+
+The format of each script is described in the documentation for
+github.com/rogpeppe/go-internal/testscript.
+
+As a special case, supporting files/directories in the goproxytest subdirectory
+will be served via a github.com/rogpeppe/go-internal/goproxytest server which
+is available to each script by default via the GOPROXY environment variable.
+The server's URL is also available via the GOPROXYTEST environment variable in
```
rogpeppe: Is this worth doing? If you want to set GOPROXY to something else, you can always set another env var yourself, can't you?

cmd/testscript/main.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+var failedRun = errors.New("failed run")
+
+type runner struct {
+	verbose bool
+}
+
+func (r runner) Skip(is ...interface{}) {
+	panic("Skip is not supported")
```
rogpeppe: maybe add a TODO comment? I think it would be good to support skip in future.

cmd/testscript/main.go
```diff
@@ -0,0 +1,249 @@
...
+
+	r := runner{
+		verbose: *fVerbose,
+	}
+
+	func() {
+		defer func() {
+			switch r := recover().(type) {
+			case nil:
+			case error:
```
rogpeppe: I think it might be clearer to check for a specific error type or value here. ISTM that the only possibilities are `failedRun` and the skip message. But then we could just make skip work if we gave it its own error value, maybe?

cmd/testscript/os.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"errors"
+	"os"
+	"runtime"
+)
+
+// os.UserHomeDir() is implemented in
+// https://go-review.googlesource.com/c/go/+/139418 so implement here for now
+func userHomeDir() (string, error) {
```
rogpeppe: This seems to be used only in tests, so maybe move it into a test file?

cmd/testscript/testdata/simple.txt
```diff
@@ -0,0 +1,41 @@
+go mod edit -replace=github.com/rogpeppe/go-internal=$GOINTERNALMODPATH
```
rogpeppe: I don't get why this is needed.

cmd/testscript/README.md
myitcv: Per our discussion, will remove and keep things simple for now.

cmd/tsrunner/main.go
```diff
@@ -0,0 +1,280 @@
...
+	Path    string  // module path
+	Version string  // module version
+	Replace *Module // replaced by this module
+	Dir     string  // directory holding files for this module, if any
+}
+
+func mainerr() (retErr error) {
+	flag.Parse()
+
+	if _, err := ioutil.TempFile("/tmp", "asdfasdf"); err != nil {
```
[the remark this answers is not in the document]
myitcv: This has been fixed in a subsequent commit.

cmd/testscript/main.go
```diff
@@ -0,0 +1,280 @@
...
+	Path    string  // module path
+	Version string  // module version
+	Replace *Module // replaced by this module
+	Dir     string  // directory holding files for this module, if any
+}
+
+func mainerr() (retErr error) {
+	flag.Parse()
+
+	if _, err := ioutil.TempFile("/tmp", "asdfasdf"); err != nil {
```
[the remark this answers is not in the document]
myitcv: This has been removed in a subsequent commit; was part of some debugging.

cmd/testscript/main.go
```diff
@@ -0,0 +1,280 @@
...
+
+type Module struct {
+	Path    string  // module path
+	Version string  // module version
+	Replace *Module // replaced by this module
+	Dir     string  // directory holding files for this module, if any
+}
+
+func mainerr() (retErr error) {
+	flag.Parse()
```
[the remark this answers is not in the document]
myitcv: Usage has been added in a subsequent commit.

cmd/testscript/main.go
```diff
@@ -0,0 +1,280 @@
...
+
+	if _, err := ioutil.TempFile("/tmp", "asdfasdf"); err != nil {
+		return fmt.Errorf("failed to create counting file: %v", err)
+	}
+
+	currProxy := os.Getenv("GOPROXY")
+
+	td, err := ioutil.TempDir("", "tsrunner_")
```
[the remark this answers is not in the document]
myitcv: Visual separation. I'm not that fussed either way on this.

cmd/testscript/main.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to create temp dir: %v", err)
+	}
+	// defer os.RemoveAll(td)
```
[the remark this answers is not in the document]
myitcv: As you suspected, this was for debugging; reenabled in a subsequent commit.

cmd/testscript/README.md
```diff
@@ -0,0 +1,111 @@
+```
+The testscript command runs github.com/rogpeppe/go-internal/testscript scripts
+in a fresh temporary work directory tree.
+
+Usage:
+    testscript [-v] files...
+
+The testscript command is designed to make it easy to create self-contained
+reproductions of command sequences. It is, in many respects, like a poor man's
```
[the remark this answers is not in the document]
myitcv: Done

cmd/testscript/README.md
```diff
@@ -0,0 +1,111 @@
...
+Docker.
+
+The format of each script is described in the documentation for
+github.com/rogpeppe/go-internal/testscript.
+
+As a special case, supporting files/directories in the goproxytest subdirectory
```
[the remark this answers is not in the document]
myitcv: Yes `goproxytest` doesn't exactly have a ring to it, other than matching the name of the package that it uses. I'll go with `.gomodproxy` if that's good with you, a blend of your suggestions above. 

And per our discussion offline I'll also only set `GOPROXY` if there are special files, and won't bother setting `GOPROXYTEST`

cmd/testscript/README.md
```diff
@@ -0,0 +1,111 @@
...
+will be served via a github.com/rogpeppe/go-internal/goproxytest server which
+is available to each script by default via the GOPROXY environment variable.
```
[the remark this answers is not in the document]
myitcv: Excellent point, thanks. Fixed.

cmd/testscript/help.go
```diff
@@ -0,0 +1,126 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"text/template"
+)
+
+func mainUsage(f io.Writer) {
+	t := template.Must(template.New("").Parse(mainHelpTemplate))
```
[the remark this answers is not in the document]
myitcv: Copy-paste from other commands... Fixed.

cmd/testscript/main.go
```diff
@@ -0,0 +1,306 @@
...
+		return 1
+	}
+	return 0
+}
+
+func mainerr() (retErr error) {
+	flag.Usage = func() {
+		mainUsage(os.Stderr)
+	}
+	flag.Parse()
```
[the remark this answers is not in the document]
myitcv: Done.

cmd/testscript/main.go
myitcv: Yes, good point. Will make the names better.

cmd/testscript/main.go
myitcv: I'll go with the standard name, thanks.

cmd/testscript/main.go
myitcv: Yup, that works.

cmd/testscript/main.go
myitcv: Yes, yet another thing I missed. Thanks for the catch.

cmd/testscript/main.go
myitcv: Yes, this thought crossed my mind. Will push up a commit with a first cut of that.

myitcv: Done, modulo some 🚲-shedding on naming. I can't come up with either a sensible name or good docs!

cmd/testscript/main.go
myitcv: > This should be shut down after the test has run, I think.

Good point. It also caught the fact that `Close()` is buggy 😄 

> Also, perhaps don't start the proxy when there are no goproxytest files?

Done.

cmd/testscript/main.go
myitcv: Per our discussion, I will implement skip support.

cmd/testscript/main.go
myitcv: Done as part of skip work.

cmd/testscript/os.go
myitcv: It's now removed so don't even need to decide where to put it.

cmd/testscript/testdata/simple.txt
myitcv: Another hangover...... removed.

cmd/testscript/main.go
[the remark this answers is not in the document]
myitcv: Removed.

cmd/testscript/README.md
```diff
@@ -0,0 +1,104 @@
...
+The testscript command is designed to make it easy to create self-contained
+reproductions of command sequences.
+
+The format of each script is described in the documentation for
+github.com/rogpeppe/go-internal/testscript.
+
+As a special case, supporting files/directories in the .gomodproxy subdirectory
+will be served via a github.com/rogpeppe/go-internal/goproxytest server which
+is available to each script via the GOPROXY environment variable. The contents
+of the .gomodproxy subdirectory are not available to the script. See the
```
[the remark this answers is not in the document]
myitcv: Done.

cmd/testscript/main.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+const (
+	// goModProxyDir is the special subdirectory in a txtar script's supporting files
+	// within which we expect to find github.com/rogpeppe/go-internal/goproxytest
+	// directories.
+	goModProxyDir = ".gomodproxy"
+)
+
+var ()
```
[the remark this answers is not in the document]
myitcv: Done.

cmd/testscript/README.md
```diff
@@ -0,0 +1,104 @@
...
+The testscript command runs github.com/rogpeppe/go-internal/testscript scripts
+in a fresh temporary work directory tree.
+
+Usage:
+    testscript [-v] files...
+
+The testscript command is designed to make it easy to create self-contained
+reproductions of command sequences.
+
+The format of each script is described in the documentation for
```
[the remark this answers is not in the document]
myitcv: Yes, that reads well and also calls out the stdin bit. Thanks

txtar/archive.go
```diff
@@ -141,3 +143,41 @@ func fixNL(data []byte) []byte {
 	d[len(data)] = '\n'
 	return d
 }
+
+// Write writes each File in an Archive to the given directory, returning any
+// errors encountered. An error is also returned in the event a file would be
+// written outside of dir.
+func Write(a *Archive, dir string) error {
```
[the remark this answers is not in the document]
myitcv: Yes I thought about this whilst on the bus home. I'll simply adapt the existing test in `txtar-x`.

cmd/testscript/main.go
```diff
@@ -0,0 +1,217 @@
...
+		} else {
+			script.Files = append(script.Files, f)
+		}
+	}
+
+	if txtar.Write(&gomodProxy, testdata); err != nil {
+		return fmt.Errorf("failed to write .gomodproxy files: %v", err)
+	}
+
+	if err := ioutil.WriteFile(filepath.Join(testdata, "script.txt"), txtar.Format(&script), 0666); err != nil {
```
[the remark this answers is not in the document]
myitcv: For stdin we could use stdin.txt

For real files, it's a bit tricky, yes. Because you might have passed two scripts, each with the same base name, but in different directories. So using the base name of the files passed in isn't perfect either. I'm tempted to go for a munged version of the name of the input files, and use that as the directory name (instead of `dirN`). Will push that up for thoughts.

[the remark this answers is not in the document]
myitcv: Actually, I've decided against the munging. It's just too messy. Instead the last line gives the file name as provided (or `<stdin>`) and tells you where it was run. I've also removed the "dir" part of `dirN`, so that 0-indexed directories within the temporary working directory are created for each file provided.

The `-work` flag (not advertised for now) prints that temporary working directory and does not remove it when the command exits (to allow you to poke around).

If the number-based naming of subdirectories proves to be annoying we can fix that down the line I guess.

cmd/testscript/main.go
```diff
@@ -0,0 +1,219 @@
...
+				err = failedRun
+			default:
+				panic(fmt.Errorf("unexpected panic: %v [%T]", err, err))
+			}
+		}()
+		testscript.RunT(r, p)
+	}()
+
+	if err != nil {
+		return fmt.Errorf("failed to run test for %v in %v\n", fileName, runDir)
```
rogpeppe: Nit: we managed to *run* the test, but it failed. How about:

    return fmt.Errorf("error running %v in %v", fileName, runDir)

?

myitcv: Good point, thanks. Fixed

gotooltest/setup.go
```diff
@@ -7,25 +7,70 @@
...
+	cmd := exec.Command("go", "list", "-f={{context.ReleaseTags}}", "runtime")
+	cmd.Stderr = &stderr
+	out, err := cmd.Output()
+	if err != nil {
+		goEnv.err = fmt.Errorf("failed to determine release tags from go command: %v\n%s", err, stderr.String())
+		return
+	}
+	goEnv.releaseTags = strings.Split(strings.Trim(strings.TrimSpace(string(out)), "[]"), " ")
+
+	getGoEnv := func(name string) (string, error) {
```
rogpeppe: How about:

    out, err := exec.Command("go", "env", "-json", "GOROOT", "GOCACHE")
    if err ...
    var vars struct {
         GOROOT, GOCACHE string
    }
    json.Unmarshal(out, &vars)
    goEnv.goroot = vars.GOROOT
    goEnv.gocache = vars.GOCACHE

?

gotooltest/setup.go
```diff
@@ -36,13 +81,14 @@ type testContext struct {
 // It checks go command can run, but not that it can build or run
 // binaries.
 func Setup(p *testscript.Params) error {
-	var c testContext
-	if err := c.init(); err != nil {
-		return err
+	goEnv.once.Do(initGoEnv)
+	if goEnv.err != nil {
```
rogpeppe: Hmm, doesn't this mean that we don't be able to use the testscript command at all if there's no go command available? I wonder if if this should fail softly and just not set the relevant environment variables if the go command isn't around.

gotooltest/setup.go
```diff
@@ -7,25 +7,70 @@
...
+	}
+	goEnv.gocache, err = getGoEnv("GOCACHE")
+	if err != nil {
+		goEnv.err = err
+		return
+	}
+
+	version := goEnv.releaseTags[len(goEnv.releaseTags)-1]
+	if !goVersionRegex.MatchString(version) {
+		goEnv.err = fmt.Errorf("invalid go version %q", version)
```
rogpeppe: You should probably return after this.
It's quite easy to make this mistake, so perhaps change initGoEnv to return an error and do:

    goEnv.once.Do(func() {
        goEnv.err = initGoEnv()
    })

?

gotooltest/setup.go
myitcv: TIL: `go env -json`. Much cleaner, thanks.

gotooltest/setup.go
myitcv: As discussed, we will only call `gotooltest.Setup` from `cmd/testscript` if `exec.LookPath("go")` succeeds. That, plus some docs to that effect, feels like the "best" thing for now.

gotooltest/setup.go
myitcv: Thanks, that's a nicer pattern.

myitcv:
@rogpeppe - made the changes discussed; added a test for the case where we call `cmd/testscript` with no `go` command available.

PR #42, opened 2019-01-07
testscript: allow non-go-test implementation

rogpeppe:


testscript/testscript.go
```diff
@@ -81,6 +81,33 @@ type Params struct {
 // RunDir runs the tests in the given directory. All files in dir with a ".txt"
 // are considered to be test files.
 func Run(t *testing.T, p Params) {
+	RunT(tshim{t}, p)
+}
+
+// T holds all the methods of the *testing.T type that
+// are used by testscript.
+type T interface {
```
myitcv: I think this interface also needs to include `Verbose() bool`, so that the later implementation can remove the check on `testing.Verbose()`.

rogpeppe: excellent point. will do.

PR #43, opened 2019-01-08
goproxytest: support fallback from txtar to reading from directory of files

myitcv:


goproxytest/proxy.go
```diff
@@ -275,6 +274,34 @@ func (srv *Server) readArchive(path, vers string) *txtar.Archive {
...
+					return err
+				}
+				if path == name && !info.IsDir() {
+					return fmt.Errorf("expected a directory root")
+				}
+				if info.IsDir() {
+					return nil
+				}
+				arpath := filepath.ToSlash(strings.TrimPrefix(path, name+string(os.PathSeparator)))
+				byts, err := ioutil.ReadFile(path)
```
rogpeppe: I might use "data" as a name here. "byts" seems a bit contrived.

myitcv: Fixed, thanks.

myitcv:
Travis seems to be having a nightmare. We have one of the lines on the board green (which covers the entire matrix) so I'm merging.

PR #45, opened 2019-01-09
cmd/testscript: add README.md that matches -help output.

myitcv:


myitcv:
Merging as this is just the addition of a markdown file (which looks good https://github.com/rogpeppe/go-internal/blob/add_testscript_command_readme/cmd/testscript/README.md)

PR #49, opened 2019-01-17
testscript: respect TestScript env PATH variable for exec command

myitcv:
Includes a partial snapshot of os/exec's platform-specific LookPath
logic, as of 0456036e28b718d215f49abe83d3c49101f8a4c7. This is largely a
copy and paste, with a type alias to os/exec.Error and an "alias" of
os/exec.ErrNotFound.

### DIFF
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/rogpeppe/go-internal/imports"
+	"github.com/rogpeppe/go-internal/internal/os/execpath"
 	"github.com/rogpeppe/go-internal/par"
 	"github.com/rogpeppe/go-internal/testenv"
 	"github.com/rogpeppe/go-internal/txtar"
@@ -167,7 +168,7 @@ type TestScript struct {
 	lineno      int               // line number currently executing
 	line        string            // line currently executing
 	env         []string          // environment list (for os/exec)
-	envMap      map[string]string // environment mapping (matches env)
+	envMap      map[string]string // environment mapping (matches env; on Windows keys are lowercase)
 	stdin       string            // standard input to next 'go' command; set by 'stdin' command.
 	stdout      string            // standard output from last 'go' command; for 'stdout' command
 	stderr      string            // standard error from last 'go' command; for 'stderr' command
@@ -220,7 +221,7 @@ func (ts *TestScript) setup() {
 	ts.envMap = make(map[string]string)
 	for _, kv := range ts.env {
 		if i := strings.Index(kv, "="); i >= 0 {
-			ts.envMap[kv[:i]] = kv[i+1:]
+			ts.envMap[envvarname(kv[:i])] = kv[i+1:]
 		}
 	}
 }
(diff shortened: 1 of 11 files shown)


testscript/testscript.go
```diff
@@ -478,6 +485,33 @@ func (ts *TestScript) execBackground(command string, args ...string) (*exec.Cmd,
 	return cmd, cmd.Start()
 }
 
+func (ts *TestScript) buildExecCmd(command string, args ...string) (*exec.Cmd, error) {
+	if filepath.Base(command) == command {
+		if lp, err := ts.lookPath(command); err != nil {
+			return nil, err
```
mvdan: Why not just let `PATH=` give the same error that `PATH=emptydir` would? You'd also make the code simpler, and not break with the `exec.Command` signature.

myitcv: Per our Slack chat; we need to eagerly use `TestScript`'s `lookPath` here, else `os/exec`'s `LookPath` will be used as part of `os/exec.Command` and then set a non-exported `error` field which is then tripped during `Start/Run` (https://github.com/golang/go/blob/006a5e7d00992cfae6ac406959512d680025f75c/src/os/exec/exec.go#L164-L177).

internal/os/exec/exec.go
```diff
@@ -0,0 +1,8 @@
+package exec
```
rogpeppe: We should probably bring along (and adapt appropriately) the tests from os/exec too (lp_test.go lp_unix_test.go lp_windows_test.go), I guess.

internal/os/exec/lp_windows.go
```diff
@@ -0,0 +1,88 @@
...
+	return "", os.ErrNotExist
+}
+
+// LookPath searches for an executable named file in the
+// directories named by the PATH environment variable.
+// If file contains a slash, it is tried directly and the PATH is not consulted.
+// LookPath also uses PATHEXT environment variable to match
+// a suitable candidate.
+// The result may be an absolute path or a path relative to the current directory.
+func LookPath(path, file string) (string, error) {
```
rogpeppe: Rather than passing in path directly, and given that different systems do different things (e.g. TEXTEXT), how about passing in a getenv function?

    // LookPath searches for an executable named file, using getenv
    // to look up environment variables. If getenv is nil, os.Getenv
    // will be used. If file contains a slash, it is tried directly and
    // getenv will not be called.
    // The result may be an absolute path or a path relative to the current directory.
    func LookPath(file string, getenv func(string) string) (string, error)

?

internal/os/exec/exec.go
rogpeppe: If we're thinking about moving this out into somewhere public, maybe "execpath" might be a good name for this package, with Look as the name of the function. Doesn't matter for now though; just a thought.

myitcv: Done. We'll see how the windows tests get on ;-)

internal/os/exec/lp_windows.go
myitcv: Done. I've also used an existing method on `TestScript` as part of this, so turned out to be a nice change.

internal/os/exec/exec.go
myitcv: Done. Nice change too! Thanks

myitcv: I had to scrap this in the end because the `os/exec` tests are quite involved and end up pulling in `internal/poll` which is not a rabbit hole we want to explore (for now) I suspect.

myitcv:
Changes made per my responses @rogpeppe.

The only additional minor change I made was to handle the fact that Windows is not case sensitive when it comes to handling environment variable names (this also picked up a bug in the `env` command)

PR #50, opened 2019-02-09
testscript: add Defer method

rogpeppe:
This allows commands and setup code to add actions that will
be performed at the end of the test. This allows us to (for example)
start a network server in the setup phase of a script and shut
it down afterwards.

Also rename the "scripts" directory to the more conventional "testdata".


myitcv:
Makes sense to me

PR #52, opened 2019-02-11
testscript: only expand a variable if it matches in the environment

myitcv:


myitcv:
Solving via a custom command, per our Slack chat.

PR #53, opened 2019-02-11
testscript: only expand second argument to cmpenv

myitcv:


testscript/testdata/cmpenv.txt
```diff
@@ -0,0 +1,8 @@
+env $=$
```
rogpeppe: ha, nice :)

testscript/testdata/cmpenv.txt
```diff
@@ -0,0 +1,8 @@
...
+cmpenv file1 file2
+
+-- file1 --
+$i
+-- file2 --
+$$i
+-- file3 --
```
rogpeppe: what's this for?

myitcv: It's not for anything. I'm an idiot 😄

PR #54, opened 2019-02-13
cmd/testscript: add support to pass through env vars with -e

myitcv:


cmd/testscript/help.go
```diff
@@ -14,7 +14,7 @@ The testscript command runs github.com/rogpeppe/go-internal/testscript scripts
 in a fresh temporary work directory tree.
 
 Usage:
-    testscript [-v] files...
+    testscript [-v] [-e VAR] files...
```
rogpeppe: maybe:

    testscript [-v] [-e VAR]... files..

?

cmd/testscript/help.go
```diff
@@ -31,6 +31,11 @@ proxy server. See the documentation for
 github.com/rogpeppe/go-internal/goproxytest for details on the format of these
 files/directories.
 
+Environment variables can be passed through to each script with the -e flag,
+where VAR is the name of the variable. Variables override testscript defined
```
rogpeppe: s/testscript defined/testscript-defined/

?

cmd/testscript/help.go
```diff
@@ -31,6 +31,11 @@ proxy server. See the documentation for
...
+values, with the exception of WORK which cannot be be overridden. The -e flag
```
rogpeppe: s/be be/be/

cmd/testscript/help.go
```diff
@@ -31,6 +31,11 @@ proxy server. See the documentation for
...
+can appear multiple times.
```
rogpeppe: s/can appear multiple times/can appear multiple times to specify multiple variables./

?

cmd/testscript/testdata/env_var.txt
```diff
@@ -0,0 +1,16 @@
+# Test passing environment variables to scripts
```
rogpeppe: Just to sanity check, how about adding a .gomodproxy entry to file.txt and check that the env var entries from that are correctly overridden (e.g. GOARCH, GOPATH)?

cmd/testscript/help.go
myitcv: Thanks.

cmd/testscript/help.go
myitcv: Thanks.

cmd/testscript/help.go
myitcv: Thanks

cmd/testscript/help.go
myitcv: Thanks

cmd/testscript/testdata/env_var.txt
myitcv: Per our chat, completely agree. 

I've actually added a more comprehensive set of tests around this. Two files, one with `go` in the `PATH`, the other without. 

The matrix of scenarios tested is then:

* with and without `go` in `PATH
* with `.gomodproxy` and without
* with `GOPROXY` set in env and without
* with `-e` set and without

This also helps to confirm that setting a custom `GOPROXY` atop a `.gomodproxy` being present in the script results in the former, i.e. the override behaviour as described. 

The code before was just wrong; the vars were applied first when (obviously) they should have been applied last.

PR #60, opened 2019-03-08
testscript: add unquote command

### LINKED ISSUE #11: proposal: txtar: ability to define txtar archives of txtar format
This currently does not work:

```
-- output.golden.txt --
-- go.mod --
module example.com/blah

-- main.go --
package main

import "fmt"

func main() {
  fmt.Println("Hello, world!")
}
```

Unsurprisingly, it results in three files, not one `output.golden.txt`

Perhaps we could use a technique similar to shell heredocs to disambiguate?

rogpeppe:
We add an unquote command to the standard set of commands,
because it's not that uncommon a requirement.

We add a `-quote` flag to txtar-savedir that will automatically
quote files if needed, adding an unquote command to the
comment section.

Future work could add a way of quoting arbitrary data (e.g. binary
data) but this should do for now.

Fixes #11.

### DIFF
--- a/testscript/doc.go
+++ b/testscript/doc.go
@@ -166,6 +166,12 @@ The predefined commands are:
 - mkdir path...
   Create the listed directories, if they do not already exists.
 
+- unquote file...
+  Rewrite each file by replacing any leading ">" characters from
+  each line. This enables a file to contain substrings that look like
+  txtar file markers.
+  See also https://godoc.org/github.com/rogpeppe/go-internal/txtar#Unquote
+
 - rm file...
   Remove the listed files or directories.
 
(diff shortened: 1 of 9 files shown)


testscript/doc.go
```diff
@@ -166,6 +166,11 @@ The predefined commands are:
 - mkdir path...
   Create the listed directories, if they do not already exists.
 
+- unquote file...
+  Rewrite each file by replacing any leading ">" characters from
```
myitcv: I wonder whether it's more intuitive/useful to reference the txtar package docs here. Because the method by which we quote is defined there, as is the unquote?

rogpeppe: I'm not sure. I think it's nice if the reader of these docs (probably someone that's trying to understand a test failure) doesn't need to go another level deeper. I'll add a note though.

PR #61, opened 2019-03-10
cmd/txtar-addmod: allow stdout destination

rogpeppe:
This makes it easier to produce inline module definitions for
inclusion in txtar files.

### DIFF
--- a/cmd/txtar-addmod/addmod.go
+++ b/cmd/txtar-addmod/addmod.go
@@ -36,6 +36,22 @@ import (
 func usage() {
 	fmt.Fprintf(os.Stderr, "usage: txtar-addmod dir path@version...\n")
 	flag.PrintDefaults()
+
+	fmt.Fprintf(os.Stderr, `
+The txtar-addmod command adds a module as a txtar archive to the
+testdata module directory as understood by the goproxytest package
+(see https://godoc.org/github.com/rogpeppe/go-internal/goproxytest).
+
+The dir argument names to directory to add the module to. If dir is "-",
+the result will instead be written to the standard output in a form
+suitable for embedding directly into a testscript txtar file, with each
+file prefixed with the ".gomodproxy" directory.
+
+In general, txtar-addmod is intended to be used only for very small
+modules - we do not want to check very large files into testdata/mod.
+
+It is acceptable to edit the archive afterward to remove or shorten files.
+`)
 	os.Exit(2)
 }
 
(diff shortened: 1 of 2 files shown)


cmd/txtar-addmod/addmod.go
```diff
@@ -36,6 +36,22 @@ import (
 func usage() {
 	fmt.Fprintf(os.Stderr, "usage: txtar-addmod dir path@version...\n")
 	flag.PrintDefaults()
+
+	fmt.Fprintf(os.Stderr, `
+The txtar-addmod command adds a module as a txtar archive to the a
+testdata module directory as understood by the goproxytest package
+(see https://godoc.org/github.com/rogpeppe/go-internal/goproxytest).
+
+The dir argument names to directory to add the module to. If dir is "-",
```
myitcv: "names _the_ directory"

myitcv: Also, I'm aware of the use of `-` as stdin, but stdout?

cmd/txtar-addmod/addmod.go
```diff
@@ -36,6 +36,22 @@ import (
...
+the result will instead be written to the standard output in a form
+suitable for embedding directly into a testscript txtar file, with each
+file prefixed with the ".gomodproxy" directory.
+
+In general, txtar-addmod is intended to be used only for very small
+modules - we do not want to check very large files into testdata/mod.
```
myitcv: I ended up hacking together a Go script that produces skeleton mod definitions. i.e. the initial `module` line followed by the `.mod` and `.info`

Might be worth providing this option via a flag?

cmd/txtar-addmod/addmod.go
```diff
@@ -133,12 +149,19 @@ func main1() int {
 		if !strings.Contains(arg, "@") {
 			title += "@" + vers
 		}
-		a.Comment = []byte(fmt.Sprintf("module %s\n\n", title))
+		dir = filepath.Clean(dir)
+		modDir := strings.Replace(path, "/", "_", -1)+"_"+vers
```
myitcv: I wonder whether we should extract this "logic"; it's repeated in `goproxytest`.

cmd/txtar-addmod/addmod.go
```diff
@@ -36,6 +36,22 @@ import (
 func usage() {
 	fmt.Fprintf(os.Stderr, "usage: txtar-addmod dir path@version...\n")
 	flag.PrintDefaults()
+
+	fmt.Fprintf(os.Stderr, `
+The txtar-addmod command adds a module as a txtar archive to the a
```
myitcv: s/to the a testdata/to the testdata/

cmd/txtar-addmod/addmod.go
rogpeppe: Yeah, it's not uncommon (the `tar` command uses that convention for example).

cmd/txtar-addmod/addmod.go
rogpeppe: Probably in another PR. I'd like to know exactly how you use this and what semantics you chose though, so I don't need to think that hard about it :)

myitcv: Yes, agreed. Doing this in another PR makes more sense.

cmd/txtar-addmod/addmod.go
rogpeppe: Yeah, probably. Perhaps not right now - this is an existing issue and it's not really a pressing one (we can't change the convention anyway, as too many existing txtar files use it)

cmd/txtar-addmod/addmod.go
rogpeppe: done

PR #68, opened 2019-04-03
testscript: provide way for Setup to pass values to commands

rogpeppe:
Currently, if Setup sets up something, there's no way for it
to pass values through to custom commands other than
strings (with Setenv). To allow this, we add a map of
arbitrary values to Env that can be accessed by custom commands.
Like context.Context, the keys are `interface{}` values to
allow external helper libraries to add their own keys without
fear of clashing with others.

### DIFF
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -39,11 +39,26 @@ var testWork = flag.Bool("testwork", false, "")
 
 // Env holds the environment to use at the start of a test script invocation.
 type Env struct {
+	// WorkDir holds the path to the root directory of the
+	// extracted files.
 	WorkDir string
-	Vars    []string
-	Cd      string
+	// Vars holds the initial set environment variables that will be passed to the
+	// testscript commands.
+	Vars []string
+	// Cd holds the initial current working directory.
+	Cd string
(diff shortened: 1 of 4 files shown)


testscript/testscript.go
```diff
@@ -39,13 +39,28 @@ var testWork = flag.Bool("testwork", false, "")
 
 // Env holds the environment to use at the start of a test script invocation.
 type Env struct {
+	// WorkDir holds the path to the root directory of the
```
mhilton: Thanks for the doc comments.

testscript/testscript_test.go
```diff
@@ -93,6 +93,11 @@ func TestScripts(t *testing.T) {
 					ts.Fatalf("setup did not see expected files; got %q want %q", setupFilenames, args)
 				}
 			},
+			"test-values": func(ts *TestScript, neg bool, args []string) {
+				if ts.Value("somekey") != 1234 {
+					ts.Fatalf("test-values did not see expected key")
```
myitcv: "... did not see expected value"

rogpeppe: done

PR #70, opened 2019-05-06
lockedfile: add Append function

Mottl:
This commit adds Append function to cmd/go/internal/lockedfile.
It has the same behaviour as Write, except that it doesn't truncate the file if it already exists.

Refer to: https://github.com/golang/go/pull/30211 (or https://go-review.googlesource.com/c/go/+/162437)

### DIFF
--- a/lockedfile/lockedfile.go
+++ b/lockedfile/lockedfile.go
@@ -120,3 +120,19 @@ func Write(name string, content io.Reader, perm os.FileMode) (err error) {
 	}
 	return err
 }
+
+// Append opens the named file (creating it with the given permissions if needed),
+// then write-locks it and appends it with the given content.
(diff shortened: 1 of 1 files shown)


myitcv:
cc @rogpeppe - note that per https://go-review.googlesource.com/c/go/+/162437 this has not been accepted upstream. So I think it comes down to working out whether we want to extend the API here not...

mvdan:
Thanks for your PR!

This function calls `OpenFile`, which is already exported, and afterwards calls `io.Copy` and `File.Close`. As far as I can tell, you can already write those eight lines of code without further API in this module. Given that this API would exist purely out of convenience for eight lines of code, I'm inclined to leave it out for now - it's already a bit of work to keep these packages up to date with usptream (see https://github.com/rogpeppe/go-internal/pull/218), so any additions or changes of our own need to be worthwhile.

PR #72, opened 2019-05-27
modfile: add tests for AddReplace and DropReplace

dmitris:
PR adds a few tests for modfile's `AddReplace` and `DropReplace` functions.


dmitris:
@rogpeppe - does the change look ok to you?

rogpeppe:
Thanks very much for suggesting these changes. In principle I am in favour, but I also wish to keep this package as close to the original `modfile` package as possible. How about adding the new tests in a separate file, say `replace_test.go`? Then we can more easily update the rest of the source files by simply copying them.

dmitris:
@rogpeppe - that's a good idea, I put the new tests in `modfile/replace_test.go`.

rogpeppe:
LGTM, thanks!

PR #77, opened 2019-08-26
ci: changes for go1.13beta1

myitcv:
Also:

* make explicit the entire Go version matrix
* make goproxytest honour the 404 behaviour expected by cmd/go in tip as
a result of https://go-review.googlesource.com/c/go/+/189517

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,22 @@
 language: go
+
 os:
   - windows
   - linux
   - osx
+
+branches:
+  only:
+    - master
+
 go:
   - "1.11.x"
   - "1.12.x"
+  - "1.13beta1"
+
 env:
   - GO111MODULE=on
+
 go_import_path: github.com/rogpeppe/go-internal
 
 # Add this before_install until we have a definitive resolution on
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -1,13 +1,22 @@
...
+
+branches:
+  only:
+    - master
+
 go:
-  - "1.11.x"
-  - "1.12.x"
+  - "1.11.13"
+  - "1.12.9"
```
mvdan: why make them explicit, and not use `.x`? this will force someone to manually update them every few weeks. you can generally rely on bugfix releases not breaking anything.

myitcv: Simply because it's then not a lottery with `gimme`. But I'm not fixed on this way or the other...

mvdan: Unless gimme or the `.x` bits give you trouble, I'd just leave it as is. I haven't had any issues for years.

One case I do pin the version is when deploying or building release binaries, or when testing at work, since we want to carefully use one Go version everywhere. But for an open source project, I think using the latest bugfix is much easier.

myitcv: Sounds like a good practical approach :)

PR #79, opened 2019-09-02
testscript: ensure that temp dir isn't a symlink

leitzler:
I ran into issues in `govim` with the testscript matcher since it uses real paths in the logs. When running macOS `ioutil.TempDir()` returns a sym-linked directory by default.

This PR ensures that the path used is a real directory by evaluating the root test temp dir that acts as base for `$WORK`.

See full description in https://github.com/myitcv/govim/issues/308

### DIFF
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -150,6 +150,14 @@ func RunT(t T, p Params) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// The temp dir returned by ioutil.TempDir might be a sym linked dir (default
+	// behaviour in macOS). That could mess up matching that includes $WORK if,
+	// for example, an external program outputs resolved paths. Evaluating the
+	// dir here will ensure consistency.
+	testTempDir, err = filepath.EvalSymlinks(testTempDir)
+	if err != nil {
+		t.Fatal(err)
+	}
 	refCount := int32(len(files))
 	for _, file := range files {
 		file := file
--- a/testscript/testdata/evalsymlink.txt
+++ b/testscript/testdata/evalsymlink.txt
@@ -1,0 +1,8 @@
+# If ioutil.TempDir returns a sym linked dir (default behaviour in macOS for example) the
+# matcher will have problems with external programs that uses the real path.
+# This script tests that $WORK is matched in a consistent way (also see #79).
+[windows] skip
+exec pwd
+stdout ^$WORK$
+exec pwd -P
+stdout ^$WORK$


myitcv:
This LGTM but we'll wait for @rogpeppe to opine. We can't test this adequately because the macOS Travis environment is not recent enough.

mvdan:
> When running macOS ioutil.TempDir() returns a sym-linked directory by default.

If that's the case, add a test to prove what this is fixing? The project is already tested on Mac via Travis.

myitcv:
> If that's the case, add a test to prove what this is fixing? The project is already tested on Mac via Travis.

I think (but @leitzler will be able to confirm) this is only surfaced on Mojave.

leitzler:
I've only tested on Mojave (10.14.6) but I don't think the current tests would catch it anyway since it's a bit of an edge case.

go-internal as well as the original go source always handles the symlink and therefor won't be affected, but as `govim` is acting from within `vim` as context where it resolves the actual path it becomes a problem.

This would end up with different paths on macOS:
```go
package main

import (
	"fmt"
	"io/ioutil"
	"os"
)

func main() {
	tmpd, err := ioutil.TempDir("", "foo")
	if err != nil {
		panic(err)
	}
	fmt.Println(tmpd)

	os.Chdir(tmpd)
	cwd, err := os.Getwd()
	if err != nil {
		panic(err)
	}
	fmt.Println(cwd)
}
```

e.g.:
```
/var/folders/xm/ls208xd174v95pgd20rn_qbh0000gn/T/foo872692457
/private/var/folders/xm/ls208xd174v95pgd20rn_qbh0000gn/T/foo872692457
```

I'll see if I can create a simple testscript that details it further.

leitzler:
Here is an example repo: https://github.com/leitzler/testscript-symlink

It's a bit hard to create a fair example since `govim` has a custom testdriver with custom commands to compare logs, but I hope this makes more sense.

testscript/testscript.go
```diff
@@ -150,6 +150,10 @@ func RunT(t T, p Params) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	testTempDir, err = filepath.EvalSymlinks(testTempDir)
```
rogpeppe: Please add a comment explaining why this is needed.

leitzler: Struggled a bit with how to explain, but now there is a comment there at least, feel free to suggest improvements :)

myitcv:
Comment looks good to me @leitzler - I think we should include the test, no? 

https://github.com/leitzler/testscript-symlink/blob/master/scripts/foo.txt

leitzler:
Good idea, I added it.

myitcv:
Thanks @leitzler 

Please can you squash the commits down and provide a descriptive commit message? 

As you can see from https://github.com/rogpeppe/go-internal/commits/master we do not use merge commits, instead relying on the commit message in the PR. When there are multiple commits in a PR (like now) GitHub simply does a concatenation of the messages... which leaves us with something less useful :)

PR #80, opened 2019-09-02
testscript: make sure work dir isn't removed when -testwork is used

### LINKED ISSUE #78: testscript: flag -testwork doesn't keep files
Log files are written to `/tmp/go-test-scriptNNNNNNNNNN` by the testscript framework.

When using the flag `-testwork` in the origin, those files are preserved:
https://github.com/golang/go/blob/25a2b98f7a1454aac0d7d3072f74613ac0446630/src/cmd/go/script_test.go#L53-L55

While the persistence decision has been moved to testscript parameters here:
https://github.com/rogpeppe/go-internal/blob/master/testscript/testscript.go#L169-L177

Shouldn't the flag be considered as well?

leitzler:
Closes #78

### DIFF
--- a/testscript/testscript_test.go
+++ b/testscript/testscript_test.go
@@ -8,9 +8,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"strconv"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


testscript/testscript_test.go
```diff
@@ -121,6 +123,30 @@ func TestScripts(t *testing.T) {
...
 
+// TestTestwork tests that using the flag -testwork will make sure the work dir isn't removed
+// after the test is done. It uses an empty testscript file that doesn't do anything.
+func TestTestwork(t *testing.T) {
+	out, err := exec.Command("go", "test", ".", "-testwork", "-v", "-run", "^TestScripts/nothing$").CombinedOutput()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	re := regexp.MustCompile(`\s+WORK=(\S+)`)
```
myitcv: We probably want to match an entire line?

leitzler: I'm not sure it adds anything since it only is a timestamp before the match.

I should however update to use FindAllStringSubmatch to make sure that the test output doesn't contain more than one line.

testscript/testscript_test.go
```diff
@@ -121,6 +123,30 @@ func TestScripts(t *testing.T) {
 	// TODO check that the temp directory has been removed.
 }
 
+// TestTestwork tests that using the flag -testwork will make sure the work dir isn't removed
+// after the test is done. It uses an empty testscript file that doesn't do anything.
+func TestTestwork(t *testing.T) {
+	out, err := exec.Command("go", "test", ".", "-testwork", "-v", "-run", "^TestScripts/nothing$").CombinedOutput()
```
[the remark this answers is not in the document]
leitzler: Oh, TIL. Fixed

testscript/testscript_test.go
leitzler: Fixed

PR #82, opened 2019-09-16
testscript: add ReadFile method

rogpeppe:
This allows custom commands to operate on the output of previous
exec commands similar to the built in cp and cmp commands.


go.mod
```diff
@@ -3,3 +3,5 @@ module github.com/rogpeppe/go-internal
 go 1.11
 
 require gopkg.in/errgo.v2 v2.1.0
+
+go 1.13
```
myitcv: Hmm how did this get added given the `go.mod` already has `go 1.11`?

rogpeppe: good point. must've been the rebase. fixed.

PR #83, opened 2019-09-29
testscript: implement UpdateScripts parameter

rogpeppe:
This makes it possible for a test to automatically update the test scripts
when output changes. Does not cover `cmpenv`.

### DIFF
--- a/testscript/testdata/testscript_update_script_not_in_archive.txt
+++ b/testscript/testdata/testscript_update_script_not_in_archive.txt
@@ -1,0 +1,12 @@
+unquote scripts/testscript.txt
+cp scripts/testscript.txt unchanged
+! testscript-update scripts
+cmp scripts/testscript.txt unchanged
+
+-- scripts/testscript.txt --
+>echo stdout right
+>cp file expect
+>cmp stdout expect
+>
+>-- file --
+>wrong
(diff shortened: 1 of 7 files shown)


testscript/testdata/testscript_update_script_not_in_archive.txt
```diff
@@ -0,0 +1,12 @@
+unquote scripts/testscript.txt
+cp scripts/testscript.txt unchanged
+testscript-update scripts
```
myitcv: Shouldn't this line need to be negated? Because the file does not exist in the archive, hence it should fall through to comparing and that should fail?

testscript/testscript.go
```diff
@@ -434,6 +452,40 @@ Script:
...
+			}
+			f.Data = data
+			found = true
+		}
+		// Sanity check.
+		if !found {
+			panic("script update file not found")
+		}
+	}
+	if err := ioutil.WriteFile(ts.file, txtar.Format(ts.archive), 0666); err != nil {
```
mpvl: Why not 644?

myitcv: I think `0666` is the "right" value here? i.e. it's the value before `umask` is applied. Hence this function is akin to `os.Create`?

testscript/testdata/testscript_update_script_not_in_archive.txt
rogpeppe: Excellent catch, sir! There was a bug in the fakeT logic (it never set the failed field). Fixed now.

testscript/testscript.go
rogpeppe: Yeah, 666 is standard.

PR #85, opened 2019-12-17
testscript: fail a test when a testscript Dir contains no scripts

myitcv:
This can happen when either the Dir does not exist, or it is empty.

### DIFF
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -152,9 +152,13 @@ func (t tshim) Verbose() bool {
 // RunT is like Run but uses an interface type instead of the concrete *testing.T
 // type to make it possible to use testscript functionality outside of go test.
 func RunT(t T, p Params) {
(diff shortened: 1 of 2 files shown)


testscript/testscript.go
```diff
@@ -126,8 +126,10 @@ func Run(t *testing.T, p Params) {
 type T interface {
 	Skip(...interface{})
 	Fatal(...interface{})
+	Fatalf(string, ...interface{})
 	Parallel()
 	Log(...interface{})
+	Logf(string, ...interface{})
```
rogpeppe: Technically adding methods here is a backwardly incompatible change, and it's not very necessary, so I think it's probably worth just using `Log` and `Fatal` with `fmt.Sprintf` instead, as we're already doing in at least one place for Fatal.

myitcv: Good point. Done

PR #86, opened 2019-12-22
testscript: document the use of GOTMPDIR

myitcv:


myitcv:
Perhaps this field would be better called `Workdir`?

testscript/doc.go
```diff
@@ -2,320 +2,321 @@
...
+//
+// 	$ WORK=/tmp/cmd-go-test-745953508/script-install_rebuild_gopath
+// 	$ cd $WORK/d1/src/p1
+// 	$ cat p1.go
+// 	package p1
+// 	import "p2"
+// 	func F() { p2.F() }
+// 	$
+//
+// The work directories will be created inside $GOTMPDIR/go-test-script*/script-$scriptname,
```
rogpeppe: i _think_ it might be better to put this text next to the the docs for `Params.TestWork` (it would be easy to miss, I think).
Also, to avoid the need to change to using `/* */` style for the whole doc comment, how about:

```
The work directories will be created inside $GOTMPDIR/go-test-script$random/script-$scriptname,
where $GOTMPDIR defaults to os.TempDir(), $random is an arbitrary random suffix and $scriptname
is the name of the testscript test file.
```

?

testscript/testscript.go
```diff
@@ -528,7 +528,7 @@ func (ts *TestScript) condition(cond string) (bool, error) {
 // abbrev abbreviates the actual work directory in the string s to the literal string "$WORK".
 func (ts *TestScript) abbrev(s string) string {
 	s = strings.Replace(s, ts.workdir, "$WORK", -1)
-	if *testWork {
+	if *testWork || ts.params.TestWork {
```
rogpeppe: oops. I suspect that the `testWork` flag was an unintentional leftover here. I'd kinda like to remove it, but probably best not.

testscript/testscript_test.go
```diff
@@ -202,6 +202,58 @@ func TestTestwork(t *testing.T) {
...
+	td, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatalf("failed to create temp dir: %v", err)
+	}
+	prev := os.Getenv("GOTMPDIR")
+	if err := os.Setenv("GOTMPDIR", td); err != nil {
+		t.Fatalf("failed to set GOTMPDIR: %v", err)
+	}
+	defer func() {
+		if err := os.Setenv("GOTMPDIR", prev); err != nil {
```
rogpeppe: I don't believe that `Setenv` can fail in this case. I wouldn't bother checking the error return, personally. (and ditto for the Setenv above).

testscript/testscript_test.go
```diff
@@ -202,6 +202,58 @@ func TestTestwork(t *testing.T) {
...
+		Dir:      filepath.Join("testdata", "nothing"),
+		TestWork: true,
+	}
+	// Run as a sub-test so that this call blocks until the sub-tests created by
+	// calling Run (which themselves call t.Parallel) complete.
+	t.Run("run tests", func(t *testing.T) {
+		Run(t, params)
+	})
+	// Verify that we have a single go-test-script-* named directory
+	dir, err := ioutil.ReadDir(td)
```
rogpeppe: I want `ioutil.Readdirnames`.

testscript/testscript_test.go
```diff
@@ -202,6 +202,58 @@ func TestTestwork(t *testing.T) {
...
+			t.Fatalf("failed to revert GOTMPDIR: %v", err)
+		}
+		os.RemoveAll(td)
+	}()
+	params := Params{
+		Dir:      filepath.Join("testdata", "nothing"),
+		TestWork: true,
+	}
+	// Run as a sub-test so that this call blocks until the sub-tests created by
+	// calling Run (which themselves call t.Parallel) complete.
```
rogpeppe: Nice.

testscript/testscript_test.go
```diff
@@ -202,6 +202,58 @@ func TestTestwork(t *testing.T) {
...
+		t.Fatalf("expected a single entry in the listing for %v: got %v %v", td, len(dir), entriesStr)
+	}
+	entry := dir[0]
+	if !strings.HasPrefix(entry.Name(), "go-test-script") {
+		t.Fatalf("unexpected name format for directory in %v: %v", td, entry.Name())
+	}
+	want := filepath.Join(td, entry.Name(), "script-nothing", "README.md")
+	if _, err := os.Stat(want); err != nil {
+		t.Fatalf("failed to find %v: %v", want, err)
+	}
```
rogpeppe: Maybe simpler: instead of the above, how about:

    files, err := filepath.Glob(filepath.Join(td, "go-test-script*", "script-nothing", "README.md"))
    if err != nil {
        t.Fatal(err)
    }
    if len(files) != 1 {
       t.Fatalf("unexpected files found for kept files; got %q", files)
    }

myitcv:
I think this is superseded by #87 but will leave open until we make a decision there.

myitcv:
Superseded by https://github.com/rogpeppe/go-internal/pull/87

PR #88, opened 2020-01-13
testscript: fix race in setupfilenames test

myitcv:
Also add go test -race ./... to CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
 install: "echo no install step required"
 
 script:
-  - go test ./...
+  - go test -race ./...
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -32,6 +32,7 @@ install: "echo no install step required"
 
 script:
   - go test ./...
```
rogpeppe: if we're running tests with `-race`, is it really worth running them without too?

myitcv: I don't know for sure, but per Slack chat let's drop the non-race line and we can read up for the guarantees or otherwise of `-race`

PR #90, opened 2020-02-23
testscript: add ${/} as a companion to ${:}

mvdan:
${:} expands to os.PathListSeparator; now ${/} expands to
os.PathSeparator.

It's useful to skip windows-vs-unixy conditions, like the ones that we
simplified in exec_path_change with HOME and PATH.

It's also useful so that 'cmpenv stdout stdout.golden' can easily
support output which contains relative file paths.

While at it, make the stdin and exec_path_change scripts work on Wine
(i.e. vanilla Windows), which has neither Go nor coreutils installed.

### DIFF
--- a/testscript/testdata/exec_path_change.txt
+++ b/testscript/testdata/exec_path_change.txt
@@ -1,17 +1,17 @@
 # If the PATH environment variable is set in the testscript.Params.Setup phase
 # or set directly within a script, exec should honour that PATH
 
-[!windows] env HOME=$WORK/home
-[windows]  env HOME=$WORK\home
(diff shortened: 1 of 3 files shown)


bcmills:
FWIW, in the `cmd/go` tests we tend to just do a regexp match on `[/\\]` (the “tent operator”) instead.

That admittedly does not work for `cmpenv`, though.

mvdan:
Yep, this is precisely to support cmpenv. There is no way to compare stdout to a `stdout.golden` file via regular expressions. I could do it with `stdout pattern...`, but then I can only check for single lines, and not a command's entire output all at once.

PR #94, opened 2020-04-13
testscript: support stdout/stderr in more builtins

mvdan:
Support these special files in the stdin, unquote, and match (grep)
commands. This is particularly useful to "pipe" output like "stdin
stdout", or to use commands that produce txtar output.

The grep one isn't super useful, as one can already do "stdout xxx"
instead of "grep xxx stdout", but this way it's more consistent and we
save some repeated lines. Because the feature is itself a bit redundant,
don't document it, to prevent confusing the user.


mvdan:
Also rog, ignore the PR title/body. The commit message is up to date.

PR #102, opened 2020-05-14
testscript: don't allow negative file modes

twpayne:
Trivial.

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -74,7 +74,7 @@ func (ts *TestScript) cmdChmod(neg bool, args []string) {
 	if len(args) != 2 {
 		ts.Fatalf("usage: chmod mode file")
 	}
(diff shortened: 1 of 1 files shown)


myitcv:
We could actually use this opportunity to refresh this method and bring it in-line with upstream:

https://github.com/golang/go/blob/881d5405402d6e8c54f83eed6216a9ed29778006/src/cmd/go/script_test.go#L456-L475

@twpayne you happy to also make those tweaks?

twpayne:
> We could actually use this opportunity to refresh this method and bring it in-line with upstream:

Done in #104.

mvdan:
@twpayne this now conflicts with master, after I merged #104 - is this PR still relevant? if so, please rebase it.

twpayne:
This PR is no longer relevant - the same change was also included in #104, which has now been merged.

mvdan:
Great! Thanks again for contributing.

PR #104, opened 2020-05-15
testscript: update chmod from golang/go

twpayne:
This pulls in the latest changes to `chmod` from upstream, as [requested by @myitcv](https://github.com/rogpeppe/go-internal/pull/102#issuecomment-629023974).

Note that this change breaks backwards compatibility. Upstream parses the permission as a regular integer, whereas testscript parses it as an octal integer, meaning that you need update your scripts to prefix permissions with a `0`.

Replaces #102.

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -71,25 +71,23 @@ func (ts *TestScript) cmdCd(neg bool, args []string) {
 }
 
 func (ts *TestScript) cmdChmod(neg bool, args []string) {
-	if len(args) != 2 {
-		ts.Fatalf("usage: chmod mode file")
-	}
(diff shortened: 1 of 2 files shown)


myitcv:
> Note that this change breaks backwards compatibility

I hadn't realised they made such a change. I think that specific change is probably a non-starter therefore. 

The other difference I spotted appeared to be backwards compatible: accepting multiple path arguments.

twpayne:
I propose to restore the existing functionality (parsing as octal). WDYT?

twpayne:
PR updated to keep parsing the mode as octal.

mvdan:
No more responses here, so I'm hitting the button.

PR #106, opened 2020-07-06
testscript: add -text argument to cmp and cmpenv

twpayne:
This PR adds a `-text` flag to testscript's builtin `cmp` and `cmpenv` commands which helps when writing cross-platform testscripts by ignoring line ending differences.

As everybody painfully knows, Windows uses different line endings to the rest of the world, which can creep in in many places:
* Depending on the user's git `core.autocrlf` setting, they may or may not appear in the txtar section of the testscript when the user checks out the testscript.
* The program being tested might emit a variety of line endings, for example using `encoding/json` [never emits Windows line endings](https://github.com/golang/go/blob/master/src/encoding/json/indent.go#L63) whereas other libraries/programs might emit Windows line endings when running on Windows.

The `-text` flag added to `cmp` and `cmpenv` in this PR sidesteps these issues by canonicalizing the output to be compared to UNIX text format before doing the comparison.

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -97,11 +97,18 @@ func (ts *TestScript) cmdCmp(neg bool, args []string) {
 		// It would be strange to say "this file can have any content except this precise byte sequence".
 		ts.Fatalf("unsupported: ! cmp")
 	}
-	if len(args) != 2 {
-		ts.Fatalf("usage: cmp file1 file2")
+
+	text := false
+	if len(args) > 0 && args[0] == "-text" {
+		text = true
+		args = args[1:]
 	}
 
-	ts.doCmdCmp(args, false)
+	if len(args) != 2 {
+		ts.Fatalf("usage: cmp [-text] file1 file2")
+	}
+
+	ts.doCmdCmp(args, text, false)
 }
 
 // cmpenv compares two files with environment variable substitution.
@@ -109,15 +116,25 @@ func (ts *TestScript) cmdCmpenv(neg bool, args []string) {
 	if neg {
 		ts.Fatalf("unsupported: ! cmpenv")
 	}
-	if len(args) != 2 {
-		ts.Fatalf("usage: cmpenv file1 file2")
+
+	text := false
+	if len(args) > 0 && args[0] == "-text" {
+		text = true
+		args = args[1:]
 	}
-	ts.doCmdCmp(args, true)
+
+	if len(args) != 2 {
+		ts.Fatalf("usage: cmpenv [-text] file1 file2")
+	}
+	ts.doCmdCmp(args, text, true)
 }
 
-func (ts *TestScript) doCmdCmp(args []string, env bool) {
+func (ts *TestScript) doCmdCmp(args []string, text, env bool) {
 	name1, name2 := args[0], args[1]
 	text1 := ts.ReadFile(name1)
+	if text {
+		text1 = toText(text1)
+	}
 
 	absName2 := ts.MkAbs(name2)
 	data, err := ioutil.ReadFile(absName2)
@@ -125,6 +142,9 @@ func (ts *TestScript) doCmdCmp(args []string, env bool) {
 	text2 := string(data)
 	if env {
 		text2 = ts.expand(text2)
+	}
+	if text {
+		text2 = toText(text2)
 	}
 	if text1 == text2 {
 		return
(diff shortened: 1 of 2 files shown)


mvdan:
I've encountered this before, and indeed I've been able to work around it via gitattributes; see https://github.com/mvdan/sh/blob/master/.gitattributes for example.

So you can fix those for an entire repo, at least, and the txtar files will always extract in a consistent way.

Beyond that, what actual problems have you encountered? I write all of my tests with Unix-like line endings, and I honestly haven't encountered a library/program that insists on using Windows line endings. As far as I know, the general consensus in Go is to not do magic things, and changing the line endings depending on GOOS feels too magical/confusing. I don't think there's a single place in the standard library that does that.

twpayne:
The actual problems I have encountered are:
1. Line endings being changed by git.
2. Line endings being changed by editors.
3. Editors adding or removing a final end-of-line to testcript files.
4. Programs (not necessarily Go programs) emitting a mix of different line endings on Windows.

Although these are mostly fixable by being extremely careful (e.g. using `.gitattributes` and ensuring that all contributors have their editors configured to behave the same way), it's simply much simpler and more practical to indicate when text comparisons (as opposed to byte-for-byte comparisons) are needed.

Therefore, I think this PR has practical value.

mvdan:
> Programs (not necessarily Go programs) emitting a mix of different line endings on Windows.

I agree that can be a tricky problem to have. You can work around it via `[windows]` and `[!windows]` conditionals, or via env vars, which is a bit cumbersome, but I've rarely encountered this to be honest.

> these are mostly fixable by being extremely careful

I personally tend to disagree. For example, the Go project used to have automatic conversion of newlines, so that any contributor to Go would properly use LF instead of CRLF, even if their editor was misconfigured. In practice, it was madness. Dozens of tests would start failing depending on what system you were on, simply because git decided that some files were plaintext and needed conversion. Similarly, some other files do need to be CRLF on master (such as bat scripts), and so they were being broken because they'd be converted to LF.

The end result is the same I suggest here - to just make git deal with chunks of bytes, and not try to interpret or convert text: https://github.com/golang/go/blob/master/.gitattributes. You do have to tell people to configure their editors correctly in this case, but honestly, that feels like a pretty simple thing for a developer to get right :)

Also, just like you can solve issue number 1 for an entire repo via `.gitattributes`, you could probably solve issues 2 and 3 for an entire repo too via an EditorConfig file: https://editorconfig.org/

twpayne:
Thanks for the detailed response :)

The project I'm using `testscript` for ([chezmoi](https://github.com/twpayne/chezmoi)) includes a number of integration tests with password manager CLIs. These are simulated in the testscripts with shell scripts on UNIX-like machines and batch scripts on Windows, for example ([source](https://github.com/twpayne/chezmoi/blob/master/testdata/scripts/secretgopass.txt)):

```
chezmoi -v secret gopass show misc/example.com
stdout examplepassword

chezmoi apply
cmp $HOME/.netrc golden/.netrc

-- bin/gopass --
#!/bin/sh

case "$*" in
"show misc/example.com")
    echo "examplepassword"
    ;;
*)
    echo "gopass: invalid command: $*"
    exit 1
esac
-- bin/gopass.cmd --
@echo off
IF "%*" == "show misc/example.com" (
    echo | set /p=examplepassword
    exit /b 0
) ELSE (
    echo gopass: invalid command: %*
    exit /b 1
)
-- home/user/.local/share/chezmoi/private_dot_netrc.tmpl --
machine example.com
login examplelogin
password {{ gopass "misc/example.com" }}
-- golden/.netrc --
machine example.com
login examplelogin
password examplepassword
```

This testscript contains a batch script, and you write above the batch scripts require CR-LF line endings, so it needs LFs converted to CR-LFs when checked out by git on Windows.

Other testscripts test my program's output, for example [this one](https://github.com/twpayne/chezmoi/blob/master/testdata/scripts/managed.txt), and my program always outputs LF line endings (e.g. using `fmt.Println`). Therefore, in contrast to the previous test, LFs must *not* be converted to CR-LFs when checked out by git on Windows.

I'm not sure if there's an easy alternative to the `-text` flag proposed in this PR. Of course, I can always maintain my own fork of this repo, but it seems a pity to do so.

mvdan:
Could you not solve this the other way around? For example, with a `unix2dos` built-in command to convert LF to CRLF, so at the top of your testscript file you could have `unix2dos bin/gopass.cmd`. That way, it would always be CRLF consistently, so it's the rough equivalent of checking a real file into git with CRLF.

twpayne:
That's a good idea, thank you. I will investigate and report back.

twpayne:
Thanks @mvdan - the approach you suggested worked. Implemented in https://github.com/twpayne/chezmoi/pull/876.

Two things that required care:
* `tar` emits LF endings on UNIX, and CRLF endings on Windows, so `unix2dos` is needed when comparing outputs.
* Some of the integration tests invoke `git`, which, by default, will transform line endings on Windows. This behavior has to be explicitly disabled.

Anyhow, it all works. Thanks again for the idea. I'll close this PR as it's no longer needed.

mvdan:
Cool, nice to hear that. I think a `unix2dos` command might be reasonable to add as a built-in, similar to how we have `chmod` - they both fill roles that txtar isn't well suited for.

PR #107, opened 2020-07-30
testscript: relax constraints around actual source in cmp with update

myitcv:
There are currently constraints on the actual source for a failed cmp
comparison when Params.UpdateScripts is true: the actual must be either
stdout or stderr.

However it's unclear why this constraint is necessary.

Indeed it's a painful constraint when the actual source in a comparison
is file written by a previous command.

Relax this constraint because, after discussion with @mvdan
and @rogpeppe, we believe it safe to do so.

Add a test for this new logic where the actual is a file that is part of
the archive.

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -131,7 +131,7 @@ func (ts *TestScript) doCmdCmp(args []string, env bool) {
 	if text1 == text2 {
 		return
 	}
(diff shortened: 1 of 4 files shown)


rogpeppe:
LGTM with one trivial typo. Thanks!

PR #109, opened 2020-08-30
testscript: support Go 1.16, move to Actions

mvdan:
https://go-review.googlesource.com/c/go/+/250977 adds a new method to
the testing.testDeps interface, so mirror that here in our no-op
implementation too.

While at it, replace Travis with GitHub Actions, which is faster and
easier to use. We still test on Linux, Mac, and Windows, and with the
two latest Go versions. Otherwise, the config is a straight port, and we
run the same commands.


.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ matrix.go-version }}
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Test
+      run: |
+        go test ./...
```
myitcv, line 19: We appear to have gained this vs Travis?

.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+        go test -race ./...
+
+    - name: Tidy
+      if: matrix.os == 'ubuntu-latest' # no need to do this everywhere
+      run: |
+        go mod tidy
+        # https://github.com/golang/go/issues/27868#issuecomment-431413621
+        go list all > /dev/null
```
myitcv, line 27: This can be dropped now, was fixed back in 1.12/1.13 from memory.

testscript/exe.go
```diff
@@ -211,3 +211,5 @@ func (nopTestDeps) WriteHeapProfile(io.Writer) error {
 	// Not needed for Go 1.10.
 	return nil
 }
+
+func (nopTestDeps) SetPanicOnExit0(bool) {}
```
myitcv, line 215: Worth a comment on why this is here, especially given we don't have a 1.16 build to break?

testscript/testdata/wait.txt
```diff
@@ -1,6 +1,10 @@
 [!exec:echo] skip
 [!exec:false] skip
 
+# TODO: the '\n' below doesn't work on Windows on Github Actions, which does
+# have coreutils like "echo" installed. Perhaps they emit CRLF?
+[windows] skip
```
myitcv, line 6: How about dropping this reliance on `echo` altogether and defining a trivial `testscript.RunMain` command?

myitcv (approved):
LGTM with a couple of suggestions.

.github/workflows/test.yml
mvdan: Ah, I added this because I do for my repos, but wasn't thinking about it too much. I tend to prefer to run without and with the race detector, because it's cheap, and because it's clearer and faster to get "normal" failures without the race detector first, in my opinion. But I can remove this if either of you disagree.

.github/workflows/test.yml
mvdan: ack, will do

testscript/exe.go
mvdan: ack, will do

testscript/testdata/wait.txt
mvdan: I leave this for a future PR, hence the TODO :) I want 1.16 support fixed ASAP, really, and this feels like low priority in comparison. Travis did not have echo installed on windows, so this was skipped there anyway.

PR #110, opened 2020-09-01
testscript: add unix2dos command

twpayne:
As [suggested by @mvdan](https://github.com/rogpeppe/go-internal/pull/106#issuecomment-684908189).

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -5,6 +5,7 @@
 package testscript
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


twpayne:
This PR doesn't include a test because:

* The functionality is fairly trivial.
* A test using a testscript would make the testscript vulnerable to the same git checkout line ending fragility that prompted the creation of the `unix2dos` command in the first place. It's nice if this repo remains encumbered with `.gitattributes` and `.editorconfig` files.

mvdan (approved):
SGTM, but I'll leave it to @rogpeppe to decide if this is something he wants to keep long-term.

testscript/cmd.go
```diff
@@ -409,6 +411,27 @@ func (ts *TestScript) cmdSymlink(neg bool, args []string) {
...
+	if len(args) < 1 {
+		ts.Fatalf("usage: unix2dos paths...")
+	}
+	for _, arg := range args {
+		filename := ts.MkAbs(arg)
+		data, err := ioutil.ReadFile(filename)
+		if err != nil {
+			ts.Fatalf("%s: %v", filename, err)
+		}
+		data = bytes.Join(bytes.Split(data, []byte{'\n'}), []byte{'\r', '\n'})
```
rogpeppe, line 428: I wonder about:
```
data = bytes.ReplaceAll(data, []byte{\n'}, []byte{'\r', '\n'})
```
which is both shorter and more efficient.

But also, I wonder if it would be better if this command was idempotent.

How about:

```
func unix2DOS(data []byte) []byte {
	var out []byte
	for scan := bufio.NewScanner(); scan.Scan(); {
		out = append(out, scan.Bytes())
		out = append(out, '\r', '\n')
	}
	return out
}
```

That way you can call unix2DOS without thinking too much.

testscript/cmd.go
```diff
@@ -409,6 +411,27 @@ func (ts *TestScript) cmdSymlink(neg bool, args []string) {
...
+		ts.Fatalf("unsupported: ! unix2dos")
+	}
+	if len(args) < 1 {
+		ts.Fatalf("usage: unix2dos paths...")
+	}
+	for _, arg := range args {
+		filename := ts.MkAbs(arg)
+		data, err := ioutil.ReadFile(filename)
+		if err != nil {
+			ts.Fatalf("%s: %v", filename, err)
```
rogpeppe, line 426:
```suggestion
			ts.Check(err)
```
The error from ReadFile already has the filename in it.

rogpeppe (commented):
Looks good, thanks! I've got a couple of minor thoughts.

testscript/cmd.go
twpayne: Done.

testscript/cmd.go
twpayne: Thanks, done.

PR #111, opened 2020-09-23
fmtsort: don't out-of-bounds panic if there's a race condition

### LINKED ISSUE #108: Port https://go-review.googlesource.com/c/go/+/191197

smasher164:
Applied upstream CL 191197 to mapelem.go. Cannot fully port to 1.11,
since MapRange started to be supported in 1.12, but we can still avoid
an index out-of-bounds panic.

Fixes #108.


myitcv (approved):
LGTM. 

In a future change we should just drop Go 1.11 support.

mvdan (approved):

PR #113, opened 2020-10-08
gotooltest: proxy the new GOMODCACHE env var too

mvdan:
(see commit message)


myitcv (approved):
LGTM. Thanks for making the 1.14 changes too.

PR #115, opened 2020-11-18
testscript: some fixes for Plan 9

fhs:
Plan 9 environment variables names can't have slash in them.
Removing this is not a big loss considering it's not available in
`cmd/go` testscripts and it's also not documented.

Update #90

### DIFF
--- a/testscript/testdata/evalsymlink.txt
+++ b/testscript/testdata/evalsymlink.txt
@@ -2,6 +2,7 @@
 # matcher will have problems with external programs that uses the real path.
 # This script tests that $WORK is matched in a consistent way (also see #79).
 [windows] skip
+[plan9] skip
 exec pwd
 stdout ^$WORK$
 exec pwd -P
--- a/testscript/testdata/exec_path_change.txt
+++ b/testscript/testdata/exec_path_change.txt
@@ -3,7 +3,7 @@
 
 [!exec:go] skip
 
-env HOME=$WORK${/}home
+env HOME=$WORK${slash}home
 [windows] env USERPROFILE=$WORK\home
 [windows] env LOCALAPPDATA=$WORK\appdata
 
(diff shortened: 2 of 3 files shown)


mvdan:
Agh. Maybe we should use a different name? Not being able to use the path separator env var in all platforms is unfortunate.

At the same time, I honestly never test on Plan9.

mvdan:
I meant a different name than `/`, but still the same name across all OSs. How about `slash`, which is still short enough? `path` is too close to `$PATH`, which already has a different meaning.

fhs:
Sorry, maybe I should've create a separate PR for the `$path` change, which is fixing a different issue on plan9. It's just a port of [an upstream CL](https://go-review.googlesource.com/126608).

As for the `/` issue, I don't have any preference on a name. Any name without `/` (including names containing multi-byte unicode characers) will work on plan9. There are many options if want something short: `$|`, `$\`, `$÷`, `$χ`, etc.

mvdan:
Ah, gotcha.

I don't think any of the other single-character env names are particularly intuitive, to be honest. I'd go with `slash`, but feel free to wait for @rogpeppe in case he disagrees.

mvdan:
We dropped this, my apologies.

The change LGTM with one nit: backwards compatibility. We need to keep `${/}` working, on platforms other than plan9, otherwise we will likely be breaking users. The new form, `${slash}`, would work on all platforms - and those projects wishing to support testing on plan9 can switch to that form.

Also, does plan9 need to expose both `$path` and `$PATH`? If not, I would expose just `$path`, like it's done for `$HOME` via `homeEnvName`.

rogpeppe:
tbh I prefer the solution chosen in this CL: https://go-review.googlesource.com/c/go/+/416554
We keep `${/}` around but don't export it, so no need for `$slash`.
Any reason that wouldn't work here?

rogpeppe:
Closing as https://go-review.googlesource.com/c/go/+/416554 has been merged and seems to address the same issue.

PR #118, opened 2021-01-18
ci: only build the master branch on push

myitcv:
Otherwise, in the case a PR is created from a branch that is pushed to
this repo, we create a workflow run for the PR and the branch itself.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,8 +1,15 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '**'
 name: Test
 jobs:
   test:
     strategy:
+      fail-fast: false
       matrix:
         go-version: [1.14.x, 1.15.x]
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,8 +1,15 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '**'
```
mvdan, line 7: I wonder, is `'*'` not enough?

mvdan (approved):
Personally, I tend to avoid this by just opening PRs from forks :) but this seems like an okay fix too.

.github/workflows/test.yml
myitcv: This was intentional. `**` is, I think, more generally correct. Because a branch could be something like `a/b/c`, in which case `*` would not match:

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

mvdan: TIL, thanks

PR #120, opened 2021-01-18
cmd/testscript: small tidy up

myitcv:


cmd/testscript/main.go
```diff
@@ -85,14 +84,26 @@ func mainerr() (retErr error) {
...
-		runDir := filepath.Join(td, strconv.Itoa(i))
+
+		// Derive a name for the directory from the basename of file, making
+		// uniq by adding a numeric suffix in the case we otherwise end
+		// up with multiple files with the same basename
+		dirName := filepath.Base(filename)
+		count := dirNames[dirName]
+		dirNames[dirName] = count + 1
+		if count != 0 {
+			dirName = fmt.Sprintf("%v%v", dirName, count)
```
mvdan, line 99: nit: use `%s%d` as it's less ambiguous, and vet will check the types for you.

mvdan: interesting that the rest of the file uses `%v` too. I personally only ever use that for interfaces or complex types like maps and structs.

mvdan (approved):

cmd/testscript/main.go
myitcv: Good point, thanks. Muscle memory more than anything.

PR #121, opened 2021-01-18
cmd/testscript: support -update flag

myitcv:


cmd/testscript/help.go
```diff
@@ -14,7 +14,7 @@ The testscript command runs github.com/rogpeppe/go-internal/testscript scripts
 in a fresh temporary work directory tree.
 
 Usage:
-    testscript [-v] [-e VAR]... files...
+    testscript [-v] [-e VAR]... [-update] files...
```
rogpeppe, line 17: Or `-u` ? :shrug:

cmd/testscript/help.go
```diff
@@ -36,6 +36,12 @@ where VAR is the name of the variable. Variables override testscript-defined
 values, with the exception of WORK which cannot be overridden. The -e flag can
 appear multiple times to specify multiple variables.
 
+The -update flag specifies that if a cmp command within a testscript fails and
+its second argument refers to a file inside the testscript file, the command
+will succeed and the testscript file will be updated to reflect the actual
+content. As such, this is the cmd/tescript equivalent of
```
rogpeppe, line 42:
```suggestion
content. As such, this is the cmd/testscript equivalent of
```

cmd/testscript/main.go
```diff
@@ -282,5 +290,16 @@ func run(runDir, filename string, verbose bool, envVars []string) error {
 		return fmt.Errorf("error running %v in %v\n", renderFilename(filename), runDir)
 	}
 
+	if update {
+		// Write the testscript archive back to the source file
+		source, err := ioutil.ReadFile(scriptFile)
+		if err != nil {
+			return fmt.Errorf("failed to read from script file %v for -update: %v", scriptFile, err)
+		}
+		if err := ioutil.WriteFile(filename, source, 0666); err != nil {
```
rogpeppe, line 299: What if we're reading from stdin and filename is `"-"` ?

cmd/testscript/main.go
```diff
@@ -282,5 +290,16 @@ func run(runDir, filename string, verbose bool, envVars []string) error {
 		return fmt.Errorf("error running %v in %v\n", renderFilename(filename), runDir)
 	}
 
+	if update {
+		// Write the testscript archive back to the source file
+		source, err := ioutil.ReadFile(scriptFile)
```
rogpeppe, line 295: Hasn't this file got the Go proxy files stripped out of it, and so won't this be losing that info from the script instead of just changing the `cmp` results? I think you probably need a merge.

cmd/testscript/testdata/update.txt
```diff
@@ -0,0 +1,19 @@
...
+
+testscript -update in.txt
+cmp in.txt res.txt
+
+-- in.txt --
+>exec printf 'hello\n'
+>cmp stdout stdout.txt
+>
+>-- stdout.txt --
+>goodbyte
```
rogpeppe, line 13: Let's add some proxy files too to make sure they're maintained.

rogpeppe (changes requested):
Thanks for doing this! AFAICS there are still a couple of issues the need fixing (or at least test cases adding for them if my assessment is flawed)...

cmd/testscript/help.go
myitcv: That works too. As ever, I'm not that fixed to any given name 👍

cmd/testscript/help.go
myitcv: Thanks for the catch. Just as I was writing this I was getting frustrated that for some reason my editor's spell checker does not seem to want to work within string literals.

cmd/testscript/main.go
myitcv: Good catch. I thought about that earlier, but then clearly got distracted.

cmd/testscript/main.go
myitcv: Excellent point. Will include a test for this too, per your comment.

cmd/testscript/main.go
myitcv: The solution I'm just about to push up retains the ability for the user to call `testscript` with a combination of `-` and a regular file. Whilst this seems a bit bizarre, no harm in retaining the ability IMO - thoughts?

myitcv:
Thanks for the review, @rogpeppe. All comments addressed, with one small question by way of reply.

cmd/testscript/help.go
```diff
@@ -14,7 +14,7 @@ The testscript command runs github.com/rogpeppe/go-internal/testscript scripts
 in a fresh temporary work directory tree.
 
 Usage:
-    testscript [-v] [-e VAR]... files...
+    testscript [-v] [-e VAR]... [-u] files...
```
rogpeppe, line 17: Please update `README.md` too, as it holds a copy of this message :\

rogpeppe (approved):
LGTM with one last thing to do (wouldn't it be nice to have only one copy of the help text?), thanks!

cmd/testscript/help.go
myitcv: Ugh. It really would be nice not to have to worry about this. Thanks for the catch!

PR #122, opened 2021-01-19
cmd/testscript: allow -e flag to pass env value

myitcv:
The -e flag for testscript currently allows the caller to specify that
environment variables are inherited from the current environment.

However this falls when the current environment does not have a value
set and you need to set a specific value. For example:

    GOBUILD=$(go env GOBUILD) testscript -e GOBUILD script.txt

Instead, allow -e to also specify a value:

    testscript -e GOBUILD=$(go env GOBUILD) script.txt

which feels more fluent/natural.

### DIFF
--- a/cmd/testscript/main.go
+++ b/cmd/testscript/main.go
@@ -273,11 +273,19 @@ func run(runDir, filename string, update bool, verbose bool, envVars []string) e
 	if len(envVars) > 0 {
 		addSetup(func(env *testscript.Env) error {
 			for _, v := range envVars {
-				if v == "WORK" {
-					// cannot override WORK
-					continue
+				varName := v
+				if i := strings.Index(v, "="); i >= 0 {
+					varName = v[:i]
+				} else {
+					v = fmt.Sprintf("%s=%s", v, os.Getenv(v))
 				}
-				env.Vars = append(env.Vars, v+"="+os.Getenv(v))
+				switch varName {
+				case "":
+					return fmt.Errorf("invalid variable name %q", varName)
+				case "WORK":
+					return fmt.Errorf("cannot override WORK variable")
+				}
+				env.Vars = append(env.Vars, v)
 			}
 			return nil
 		})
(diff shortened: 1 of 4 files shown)


cmd/testscript/main.go
```diff
@@ -273,11 +273,21 @@ func run(runDir, filename string, update bool, verbose bool, envVars []string) e
...
-				if v == "WORK" {
+				varName := v
+				if i := strings.Index(v, "="); i >= 0 {
+					varName = v[:i]
+				} else {
+					v = fmt.Sprintf("%s=%s", v, os.Getenv(v))
+				}
+				switch varName {
+				case "":
+					// Nonense variable
```
rogpeppe, line 284: Maybe this would be better as an error, as it's almost certainly a mistake.

cmd/testscript/main.go
```diff
@@ -273,11 +273,21 @@ func run(runDir, filename string, update bool, verbose bool, envVars []string) e
 	if len(envVars) > 0 {
 		addSetup(func(env *testscript.Env) error {
 			for _, v := range envVars {
-				if v == "WORK" {
+				varName := v
+				if i := strings.Index(v, "="); i >= 0 {
+					varName = v[:i]
+				} else {
+					v = fmt.Sprintf("%s=%s", v, os.Getenv(v))
```
rogpeppe, line 280:
```suggestion
					v += "="+os.Getenv(v)
```
?

cmd/testscript/main.go
```diff
@@ -273,11 +273,21 @@ func run(runDir, filename string, update bool, verbose bool, envVars []string) e
...
+				}
+				switch varName {
+				case "":
+					// Nonense variable
+					continue
+				case "WORK":
```
rogpeppe, line 286: tbh, this should probably be an error too.

rogpeppe (approved):
LGTM with a couple of minor suggestions, thanks!

cmd/testscript/main.go
myitcv: Yep, fair point.

cmd/testscript/main.go
myitcv: Yep, fair point.

cmd/testscript/main.go
myitcv: Done.

PR #125, opened 2021-01-21
cmd/testscript: make -work actually work

myitcv:
Turns out that we were only printing the cmd/testscript created work
directory for an argument. We also need to print (and preserve) the
working directory that is created by testscript itself when running the
script.

Combine this with a bit of light refactoring to improve readability.

### DIFF
--- a/cmd/testscript/main.go
+++ b/cmd/testscript/main.go
@@ -99,6 +99,13 @@ func mainerr() (retErr error) {
 		return fmt.Errorf("cannot use -u when reading from stdin")
 	}
 
+	tr := testRunner{
+		update:   *fUpdate,
+		verbose:  *fVerbose,
+		env:      envVars.vals,
+		testWork: *fWork,
+	}
+
 	dirNames := make(map[string]int)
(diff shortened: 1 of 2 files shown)


cmd/testscript/main.go
```diff
@@ -252,6 +224,13 @@ func run(runDir, filename string, update bool, verbose bool, envVars []string) e
 		}
 	}
 
+	if tr.testWork {
+		addSetup(func(env *testscript.Env) error {
+			fmt.Fprintf(os.Stderr, "temporary work directory for %s: %v\n", renderFilename(filename), env.WorkDir)
```
mvdan, line 229: nit: %s for WorkDir

mvdan (approved):
nice work :wink:

cmd/testscript/main.go
myitcv: I will break myself of this habit one day. Thank you

myitcv:
> nice work 😉

Thanks. Together we can work this out.

PR #134, opened 2021-03-01
testscript: explore addition of io/fs.FS to Params

myitcv:
DO NOT SUBMIT

### DIFF
--- a/testscript/params_go1.16.go
+++ b/testscript/params_go1.16.go
@@ -1,0 +1,78 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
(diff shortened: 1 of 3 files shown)


myitcv:
The build failures are as a result of https://github.com/golang/go/issues/40067#issuecomment-782075833. So whilst not directly caused by this change, this change does exacerbate the situation.

myitcv:
Per a quick chat with @mvdan, perhaps a better approach here is to walk the entire `io/fs.FS` for `.txt` files, rather than simply globbing at the root. This would work well, because the caller would be entirely in control over what is exposed via the `FS`.

rogpeppe:
This has been sitting around long enough that I think it can be closed. It's not clear exactly what need is being addressed here. Feel free to reopen if needed.

PR #135, opened 2021-03-12
testscript: create a hidden temp directory by default

myitcv:
When running a testscript, we currently create a temporary directory in
$WORK/tmp. This is problematic because it interacts badly with cmd/go
(which is used by go/packages) trying to resolve the ./... pattern from
$WORK.

Instead establish a temporary directory in workdir, but use a prefix
that ensures this directory will not be walked when resolving the ./...
pattern from workdir.

### DIFF
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -289,13 +289,23 @@ type backgroundCmd struct {
 // It returns the comment section of the txtar archive.
 func (ts *TestScript) setup() string {
 	ts.workdir = filepath.Join(ts.testTempDir, "script-"+ts.name)
-	ts.Check(os.MkdirAll(filepath.Join(ts.workdir, "tmp"), 0777))
+
+	// Establish a temporary directory in workdir, but use a prefix that ensures
+	// this directory will not be walked when resolving the ./... pattern from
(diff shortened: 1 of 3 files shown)


testscript/testscript.go
```diff
@@ -289,13 +289,23 @@ type backgroundCmd struct {
...
 	ts.workdir = filepath.Join(ts.testTempDir, "script-"+ts.name)
-	ts.Check(os.MkdirAll(filepath.Join(ts.workdir, "tmp"), 0777))
+
+	// Establish a temporary directory in workdir, but use a prefix that ensures
+	// this directory will not be walked when resolving the ./... pattern from
+	// workdir. This is important because when resolving a ./... pattern, cmd/go
+	// (which is used by go/packages) creates temporary build files and
+	// directories. This can, and does, therefore interfere with the ./...
+	// pattern when used from workdir and can lead to race conditions within
+	// cmd/go as it walks directories to match the ./... pattern
```
mvdan, line 299: nit: period ending the sentence

mvdan (approved):

testscript/testscript.go
myitcv: Done

PR #137, opened 2021-03-29
testscript: workaround quardratic behaviour of github.com/pkg/diff

myitcv:
The current version of github.com/pkg/diff.Diff has quadratic behaviour.
This means that when we attempt a diff between relatively modest sized
files, it's easy to find yourself out of memory.

Therefore, when we see a diff between two large files (large defined in
terms of number of lines), tersely report that as if the two were binary
files, i.e. do not try to render the diff.

When github.com/pkg/diff or similar supports a linear algorithm:

https://github.com/golang/go/issues/45200#issuecomment-806171109

we can revert this change.


testscript/cmd.go
```diff
@@ -141,6 +141,12 @@ func (ts *TestScript) doCmdCmp(args []string, env bool) {
 		// update the script.
 	}
 
+	const maxLineDiff = 1_000_000
```
mvdan, line 144: Document how you arrived at this number? For example:

```
pkg/diff is quadratic at the moment.
If the product of the number of lines in the inputs is too large,
don't call pkg.Diff at all as it might take tons of memory or time.
We found one million to be reasonable for an average laptop.
```

mvdan (approved):

testscript/cmd.go
myitcv: Good point. I was being lazy given this is a temporary change. Done.

PR #138, opened 2021-03-29
ci: handle rename of master to main

myitcv:


mvdan:
Hmm, but we didn't rename the default branch?

myitcv:
Yes, absolutely. This issue was created before GitHub added a whole load of features to make it easier to rename the default branch (which was a blessed relief)

PR #140, opened 2021-05-02
testscript: support writing to stdio from inside a custom command

### LINKED ISSUE #139: Allow custom commands to write to stdout and stderr
Hey there, I've had a couple cases where it'd be nice to be able to write to stdout and stderr from inside a custom command. 

For example:

```go
func Test(t *testing.T) {
	p := testscript.Params{
		Cmds: map[string]func(ts *testscript.TestScript, neg bool, args []string){
			"render": render,
		},
	}
	testscript.Run(t, p)
}

func render(ts *testscript.TestScript, neg bool, args []string) {
    html, err := render.Render(ts.MkAbs(args))
    ts.Check(err)
    _ = html
}
```

```txt
render 'index.gohtml'

-- index.gohtml --
<html>
  <body>
    <h1>{{.}}</h1>
  </body>
</html>
```

Unfortunately I haven't found a way to use the command's results within the test. The workaround I use now is to write a file and then read from it.

What I'd like to be able to do is write a testscript like this:

```txt
render 'index.gohtml'
stdout '<h1>'

-- index.gohtml --
<html>
  <body>
    <h1>{{.}}</h1>
  </body>
</html>
```

And the failure case:

```txt
! render 'index.gohtml'
stderr 'bad HTML'

-- index.gohtml --
<html
```

Thanks for extracting and improving this package!

matthewmueller:
This PR allows you to write to stderr and stdout from within a custom command. For example,

```go
func Test(t *testing.T) {
	p := testscript.Params{
		Cmds: map[string]func(ts *testscript.TestScript, neg bool, args []string){
			"render": render,
		},
	}
	testscript.Run(t, p)
}

func render(ts *testscript.TestScript, neg bool, args []string) {
    html, err := render.Render(ts.MkAbs(args))
    if err != nil {
      if neg {
        ts.Stdout("")
        ts.Stderr(err.Error())
        return
      }
      ts.Fatalf("Error rendering")
    }
    ts.Stdout(html)
    ts.Stderr("")
}
```

Allowing you to have a test like this:

```txt
render 'index.gohtml'
stdout '<h1>'
! stderr .

-- index.gohtml --
<html>
  <body>
    <h1>{{.}}</h1>
  </body>
</html>
```

And 

```txt
! render 'index.gohtml'
stderr 'bad HTML'
! stdout .

-- index.gohtml --
<html
```

Closes #139

### DIFF
--- a/testscript/testdata/commandstdio.txt
+++ b/testscript/testdata/commandstdio.txt
@@ -1,0 +1,24 @@
+[!exec:cat] stop
+
+# Add to stdout
+exec cat hello.text
+! stderr .
+stdout 'hello world'
+
+# Reverse stdout
+reverse 'hello world'
+! stderr .
+stdout 'dlrow olleh'
+
+# Test that we didn't mess anything up
+exec cat hello.text
+! stderr .
+stdout 'hello world'
+
+# Test stderr
+! reverse ''
+! stdout .
+stderr 'reverse can''t be blank'
+
+-- hello.text --
+hello world
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -779,6 +779,22 @@ func (ts *TestScript) Getenv(key string) string {
 	return ts.envMap[envvarname(key)]
 }
 
+// Stdout sets stdout
+func (ts *TestScript) Stdout(stdout string) {
+	if ts.stdout != "" {
+		ts.Logf("[stdout]\n%s", stdout)
+	}
+	ts.stdout = stdout
+}
+
+// Stderr sets stderr
+func (ts *TestScript) Stderr(stderr string) {
+	if ts.stderr != "" {
+		ts.Logf("[stderr]\n%s", stderr)
+	}
+	ts.stderr = stderr
+}
+
 // parse parses a single line as a list of space-separated arguments
 // subject to environment variable expansion (but not resplitting).
 // Single quotes around text disable splitting and expansion.
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -781,7 +781,7 @@ func (ts *TestScript) Getenv(key string) string {
 
 // Stdout sets stdout
 func (ts *TestScript) Stdout(stdout string) {
-	if ts.stdout != "" {
+	if stdout != "" {
 		ts.Logf("[stdout]\n%s", stdout)
 	}
 	ts.stdout = stdout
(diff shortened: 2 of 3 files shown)


matthewmueller:
One thing that somewhat surprised me during testing was the need to do:

```go
ts.Stderr("")
```

This was to make sure the previous command's stderr didn't linger. I could also adjust the API to:

```
ts.Stdio(stdout, stderr)
```

This would be slightly more cumbersome but ensure you don't forget to also set stderr.

mvdan:
Sorry for the silence; this package is effectively maintained by three people, so sometimes none of us takes the initiative with new features like this :)

I agree that this is a limitation of the current API, and with @rogpeppe we've talked about this in the past. So I find it to be an entirely reasonable feature request to want to write to stdout - moving the command to the top level via `RunMain` is not a great solution, because it requires having a `func TestMain` (i.e. run as part of `go test`), and it also means running as a separate process.

As for the patch itself: first, I wonder if we need `Stderr` at all. Intuitively, I only want my builtins to write to stderr when they encounter an error and fail. You can already do this via https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript#TestScript.Fatalf, and I argue that is a better API anyway - it also makes the builtin result in a failure status code.

I also think the behavior of `Stdout` might confuse some people. It replaces the entire output at each call, but it allow you to perform repeated calls and will happily log each one of those strings.

I think we have two options:

1) Only allow up to one call to `Stdout`. Repeated calls result in a fatal error, as it's a bug in the user's code.

2) Allow any number of calls, e.g. via one of these two APIs:

```
// easier to use for simple stuff and mirrors Fatalf, but no io.Writer
func (ts *TestScript) Printf(format string, args ...interface{})

// my personal preference:
// integrates with any io.Writer API nicely, and simple use cases can be fmt.Fprintf(ts.Stdout(), ...)
func (ts *TestScript) Stdout() io.Writer
```

mvdan:
Also FYI @twpayne as we quite literally talked about this last week at the Go meetup in Zurich - how you can't move a top-level `RunMain` command to a builtin because builtins have no stdout.

mvdan:
Thanks for your response! I'll leave this open for now then, to see if someone else needs this. I personally don't have a need for it, so I don't have an urge to get it done, but I also don't think we need to close the issue and PR right away.

matthewmueller:
Deciding to close this one to cleanup my open PRs view a bit. If someone needs this feel free to use this PR as a base!

PR #145, opened 2021-09-22
testscript: add support for Go master aka 1.18

mvdan:
(see commit message)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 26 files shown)


myitcv (approved):
LGTM although I'll declare I'm not too familiar with the fuzzy stuff

rogpeppe (approved):
LGTM, thanks!

mvdan:
@rogpeppe @myitcv test-gotip is now added. It takes about 3m, compared to just under 1m for the stable version. That is to be expected, as building from source takes between a minute and two on average hardware.

myitcv:
Still LGTM. Thank you for going the extra mile 😄

PR #148, opened 2021-12-05
testscript: print background command output on failure

rogpeppe:
When a script has failed, the output of the commands running in the
background can be useful to see, so display it then and also in verbose
mode.

### DIFF
--- a/cmd/testscript/main.go
+++ b/cmd/testscript/main.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


myitcv:
Interesting. I've needed something similar in the past. Can we add a test to cover this?

rogpeppe:
I thought about adding a test but couldn't think of a straightforward way to do it without checking the whole testing infrastructure we've got going here. Any suggestions?

rogpeppe:
I'm going to merge this for now.

PR #149, opened 2021-12-15
testscript: add go:version condition

twpayne:
Changes in Go's standard library can affect program behavior. For example, Go 1.18's `text/template` changes the behavior of the `and` and `or` to lazily evaluate their arguments. This can impact application behavior.

This PR adds a `go:version` condition that allows testscripts to be conditional on the Go version being used.

### DIFF
--- a/testscript/doc.go
+++ b/testscript/doc.go
@@ -105,6 +105,7 @@ should only run when the condition is satisfied. The predefined conditions are:
  - [link] for whether the OS has hard link support
  - [symlink] for whether the OS has symbolic link support
  - [exec:prog] for whether prog is available for execution (found by exec.LookPath)
+ - [go:version] for whether the Go version is at least the given version
 
 A condition can be negated: [!short] means to run the rest of the line
 when testing.Short() is false.
--- a/testscript/testdata/goversion.txt
+++ b/testscript/testdata/goversion.txt
@@ -1,0 +1,8 @@
+[!exec:sh] skip 'sh not found in $PATH'
+
+# test that the go: condition works for minimum supported versions of Go
+[go:1.16] exec sh -c 'exit 0'
+[go:1.17.1] exec sh -c 'exit 0'
+
+# test that the go:condition works for future versions of Go
+[go:2] exec sh -c 'exit 1'
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -26,6 +26,7 @@ import (
 	"github.com/rogpeppe/go-internal/imports"
 	"github.com/rogpeppe/go-internal/internal/os/execpath"
 	"github.com/rogpeppe/go-internal/par"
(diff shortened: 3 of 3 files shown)


mvdan:
Are you aware that you can use build tags for this, such as `[go1.18]` or `[!go1.17]`?

twpayne:
> Are you aware that you can use build tags for this, such as `[go1.18]` or `[!go1.17]`?

No, I was not aware. Are you sure that this is the case? The word "tag" does appear anywhere in the `testscript/` subdirectory of this repo, and when I try a simple test I get an "unknown condition" error:

```console
$ go test ./testscript
--- FAIL: TestScripts (0.00s)
    --- FAIL: TestScripts/goversion (0.00s)
        testscript.go:397: 
            # test that the go: condition works for minimum supported versions of Go (0.000s)
            > [go1.16] exec sh -c 'exit 0'
            FAIL: testdata/goversion.txt:4: unknown condition "go1.16"
            
FAIL
FAIL    github.com/rogpeppe/go-internal/testscript      0.846s
FAIL
```

mvdan:
Ah, I was slightly wrong. We do already support this via conditions, but only if you opt into gotooltest: https://github.com/rogpeppe/go-internal/blob/115ce09fd6b421993aafa8cdae0171429e9bd2c4/gotooltest/setup.go#L115-L122

mvdan:
So I think you could classify this as "gotooltest is not particularly well documented" :) Though, then again, its code is barely 150 lines.

twpayne:
Ah, thank you. In my case I'm looking to test an application built with Go rather than the `go` command itself, so I don't think using `gotooltest` is appropriate.

Using build tags is a much nicer solution that what's in this PR. What do you think about replacing this PR with a PR that implements the `tag:foo` condition that returns true iff the build tag `foo` is set?

mvdan:
Yeah, perhaps using gotooltest would be a bit overkill.

An alternative might be to promote those conditionals to the testscript package. I really don't see why they are specific to testing Go tools. You might want to write conditionals based on the Go version for the standard library changes, like you say. That seems to me like a smaller change overall, since we wouldn't end up with two ways to write the same conditional.

I'm also not sure your proposed fix is right, because it seems to only look at ReleaseTags, so I'm pretty sure it won't work for any arbitrary build tag like `-tags=foobar`.

twpayne:
> I'm also not sure your proposed fix is right, because it seems to only look at ReleaseTags, so I'm pretty sure it won't work for any arbitrary build tag like `-tags=foobar`.

Are the actual build tags like `foobar` available to runtime Go code? I didn't even know about `go/build.Default.ReleaseTags` until I saw it :) In the short term I'll update the doc for #150 to mention release tags only.

twpayne:
Closing in favor of #150.

PR #150, opened 2021-12-15
testscript: add build:tag condition

twpayne:
As discussed with @mvdan in #149.


twpayne:
The `build:` prefix is needed to distinguish between set and not-set tags. If the syntax is only `tag` (e.g. `[go1.19]`) then it is not possible to distinguish between a not-set tag and an unknown condition.

testscript/doc.go
```diff
@@ -105,6 +105,8 @@ should only run when the condition is satisfied. The predefined conditions are:
  - [link] for whether the OS has hard link support
  - [symlink] for whether the OS has symbolic link support
  - [exec:prog] for whether prog is available for execution (found by exec.LookPath)
+ - [build:tag] for whether the release build tag is set; user-defined build tags
+   are not currently supported
```
rogpeppe, line 109:
```suggestion
   are not currently supported. Note that the "build:" prefix
   distinguishes this set of conditions from the conditions recognized
   by the gotooltest package - that package has no "build:" prefix and
   guards based on whether the go tool supports a given tag, but the
   "build:" tag guards based on whether the binary that's imported the
   testscript package supports the tag.

```

rogpeppe (commented):
Sorry for the slow review. This LGTM modulo a couple of minor suggestions, and I think it supercedes #149, right? Thanks for proposing it!

twpayne:
Thanks for the review and zero stress about the review time - you're giving the gift of open source so there are no expectations or obligations.

I've applied your suggested changes and rebased on the latest `master`, so this should be good to merge.

twpayne:
Anything else needed here?

twpayne:
#154 implements @mvdan's suggestions, superseding this PR.

twpayne:
Closing in favor of #154.

PR #152, opened 2022-01-12
testscript: support named background commands

rogpeppe:
This lets us wait for an individual background command rather
than all of them at once.


mvdan:
I don't have strong feelings about this, but can't object either. SGTM

PR #154, opened 2022-03-02
testscript: add Go compiler and version conditions

twpayne:
As suggested by @mvdan in https://github.com/rogpeppe/go-internal/pull/150#issuecomment-1055495507. Replaces #150.

### DIFF
--- a/gotooltest/setup.go
+++ b/gotooltest/setup.go
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	goVersionRegex = regexp.MustCompile(`^go([1-9][0-9]*)\.(0|[1-9][0-9]*)$`)
+	goVersionRegex = regexp.MustCompile(`^go([1-9][0-9]*)\.([1-9][0-9]*)$`)
 
 	goEnv struct {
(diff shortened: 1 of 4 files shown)


gotooltest/setup.go
```diff
@@ -112,7 +109,7 @@ func Setup(p *testscript.Params) error {
 			// that will be used.
 			return cond == runtime.Compiler, nil
 		}
-		if goVersionRegex.MatchString(cond) {
+		if testscript.GoVersionRegex.MatchString(cond) {
```
mvdan, line 112: It seems to me like gotooltest should remove the code to deal with these gc/gccgo and goX.Y conditions, as they are now handled by the main gotooltest package. Then you also remove the need to export GoVersionRegex, which isn't great as it's just an internal detail :)

mvdan (approved):
SGTM as long as we remove the now unnecessary code in gotooltest.

gotooltest/setup.go
twpayne: Thanks for the feedback, done.

PR #155, opened 2022-03-09
testscript: remove errgo dependency

rogpeppe:
There's no need to use it and every dependency lost is good.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,4 @@ module github.com/rogpeppe/go-internal
 
 go 1.16
 
-require (
-	github.com/pkg/diff v0.0.0-20210226163009-20ebb0f2a09e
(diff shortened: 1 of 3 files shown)


mvdan (approved):

thepudds:
This is great - thanks!

FWIW, there are a couple places I would like to use go-internal but I’m not due to people’s sensitivities around dependencies (coupled with the broader issue of it’s not easy or broadly known how to differentiate test-only dependencies…)

thepudds:
That’s also partly what caused me to investigate https://github.com/pkg/diff/issues/26#issuecomment-1045170532 (along with a nerd-sniping tweet from @myitcv)…

… but I am progressively getting more off-topic – mainly just wanted to say thanks!

PR #158, opened 2022-06-10
drop Go 1.16, add Go 1.18

mvdan:
(see commit message)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,5 @@
 module github.com/rogpeppe/go-internal
 
-go 1.16
+go 1.17
 
 require github.com/pkg/diff v0.0.0-20210226163009-20ebb0f2a09e
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,17 +11,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.16.x, 1.17.x]
+        go-version: [1.17.x, 1.18.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
     - name: Install Go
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -1,5 +1,5 @@
 module github.com/rogpeppe/go-internal
 
-go 1.16
+go 1.17
```
rogpeppe, line 3: Could we leave support for go1.16 until we actually require go1.17 features?
I suspect there are a bunch of people still using 1.16 out there.

rogpeppe (approved):
LGTM with one thought. WDYT?

go.mod
mvdan: Is that a policy we want to enforce going forward? We didn't support Go 1.15 when 1.17 was the latest version, so what you say would be a change - supporting 3 major versions rather than 2. I personally think it's unnecessary at this point, given that 1.18 has been out for three months and 1.16.x has been unsupported upstream for a while too :)

Worth noting that bumping this version here doesn't preclude building on 1.16; in fact it should continue to work until we actually do start using new features/APIs. But more than anything, it's a signal that we only test and explicitly support 1.16 or later.

rogpeppe: I reckon that it's nicer to keep support for a given version until we need to drop it tbh. Is there any significant cost to us doing so?

mvdan: No immediate cost, but CI does get slower the more versions we support, and we do have shenanigans like dealing with the signature of `testing.MainStart` for example. We have 6 CI jobs and we'd likely jump to 9.

If we want to change the supported version policy that's fine with me, I just wasn't aware that was your intention. Either way, I'll leave it to you to wrap up this PR however you see fit, because I don't have strong opinions either way :)

PR #159, opened 2022-06-13
all: add support for txtar extension and prefer it

### LINKED ISSUE #126: testscript: add support for the txtar file extension
Currently, [`testscript`](https://github.com/rogpeppe/go-internal/blob/dc4b49510d96256f881880b9802536a6a70edfcc/testscript/testscript.go#L194) is hard-coded to only look for `.txt` files.

This is unfortunate, because `.txtar` is a more descriptive extension that would better lend itself to automatic syntax highlighting (on GitHub) etc. 

Opening this issue to capture the options for changing/extending this:

1. switch the default to search for `.txt` and `.txtar`
2. make it configurable via `testscript.Params`
3. ...

mvdan:
(see commit message)

Fixes #126.

### DIFF
--- a/goproxytest/proxy.go
+++ b/goproxytest/proxy.go
@@ -7,9 +7,9 @@ Package goproxytest serves Go modules from a proxy server designed to run on
 localhost during tests, both to make tests avoid requiring specific network
 servers and also to make them significantly faster.
 
-Each module archive is either a file named path_vers.txt or a directory named
-path_vers, where slashes in path have been replaced with underscores. The
-archive or directory must contain two files ".info" and ".mod", to be served as
+Each module archive is either a file named path_vers.txtar or path_vers.txt, or
+a directory named path_vers, where slashes in path have been replaced with underscores.
+The archive or directory must contain two files ".info" and ".mod", to be served as
 the info and mod files in the proxy protocol (see
 https://research.swtch.com/vgo-module).  The remaining files are served as the
 content of the module zip file.  The path@vers prefix required of files in the
@@ -96,10 +96,15 @@ func (srv *Server) readModList() error {
 	}
 	for _, info := range infos {
 		name := info.Name()
-		if !strings.HasSuffix(name, ".txt") && !info.IsDir() {
+		switch {
+		case strings.HasSuffix(name, ".txt"):
+			name = strings.TrimSuffix(name, ".txt")
+		case strings.HasSuffix(name, ".txtar"):
+			name = strings.TrimSuffix(name, ".txtar")
+		case info.IsDir():
+		default:
 			continue
 		}
-		name = strings.TrimSuffix(name, ".txt")
 		i := strings.LastIndex(name, "_v")
 		if i < 0 {
 			continue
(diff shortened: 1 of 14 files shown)


goproxytest/proxy.go
```diff
@@ -96,10 +96,15 @@ func (srv *Server) readModList() error {
 	}
 	for _, info := range infos {
 		name := info.Name()
-		if !strings.HasSuffix(name, ".txt") && !info.IsDir() {
+		switch {
+		case strings.HasSuffix(name, ".txt"):
+			name = strings.TrimSuffix(name, ".txt")
+		case strings.HasSuffix(name, ".txtar"):
+			name = strings.TrimSuffix(name, ".txtar")
+		case info.IsDir():
```
rogpeppe, line 104: Why are we allowing a directory when we weren't before? Maybe worth a comment?

testscript/testscript.go
```diff
@@ -200,13 +200,21 @@ func (t tshim) Verbose() bool {
 // RunT is like Run but uses an interface type instead of the concrete *testing.T
 // type to make it possible to use testscript functionality outside of go test.
 func RunT(t T, p Params) {
-	glob := filepath.Join(p.Dir, "*.txt")
-	files, err := filepath.Glob(glob)
-	if err != nil {
+	entries, err := os.ReadDir(p.Dir)
+	if os.IsNotExist(err) {
+	} else if err != nil {
```
rogpeppe, line 205:
```suggestion
	if err != nil && !os.IsNotExist(err) {
```
?

rogpeppe (approved):
LGTM with a couple of minor suggestions, thanks!

goproxytest/proxy.go
mvdan: we did continue the loop in the case of `info.IsDir()` before; see the deleted line. I essentially inverted the `!foo && !bar` with a `break` into a `foo || bar` with an `else` (the `default` case) which breaks.

one logical difference is that the old code would accept `module_version.txt/` as a directory, happily stripping the `.txt` suffix. I think that was a coding mistake, because that feature was never documented - the docs only refer to `module_version.txt` as a file, and `module_version/` as a directory.

testscript/testscript.go
mvdan: The old globbing code would return zero matches due to a missing directory as `nil, nil`, but `os.ReadDir` errors - so to match the old behavior, I ignore "does not exist" errors for the directory. This makes one of the tests happy, and IMO it's nicer to say "no txtar files found in directory" rather than a platform specific "open directory: not found".

mvdan: Added the comment. I kept the logic as it is now, because I can more clearly add the comment about IsNotExist.

mvdan:
@rogpeppe are you OK with me merging this? :)

goproxytest/proxy.go
rogpeppe: Sorry, I misread the previous logic. All LGTM.

rogpeppe (approved):

PR #160, opened 2022-06-21
testscript: exit 0 in runCoverSubcommand if -help is used

FiloSottile:
flag.ExitOnError will exit 0 for -h/-help, while the flag.PanicOnError
override in runCoverSubcommand was always exiting 2, making the
success/failure of running a command with -hhelp depend on whether
coverage is active.


mvdan (approved):
SGTM. We probably inherited or wrote this code before https://github.com/golang/go/issues/37533 landed.

PR #164, opened 2022-06-28
testscript: add Params.RequireExplicitExec

### LINKED ISSUE #163: suggest using `exec foo` for top-level commands rather than `foo` for consistency
At https://github.com/cue-lang/cue we use test scripts for our integration tests, as well as for users providing reproducers for bugs.

There's one unfortunate inconsistency: the former use `cue`, and the latter use `exec cue`.

I initially thought that we should use `cue` everywhere, by adding a new flag to `cmd/testscript` like `-x cue`, so that a command "passes through" to executing a program from `$PATH`. Then, we would use `cue` everywhere rather than `exec cue`. However, then bug report reproducers aren't truly standalone, as one needs to remember to use the flag.

@rogpeppe suggests that we should use `exec cue` everywhere instead. This already works in today's `testscript.RunMain`, as top-level commands are put in a temporary directory and appended to `$PATH`. We simply prefer `cue` rather than `exec cue` since we're used to how it was before, where `exec cue` might fail to find the program if one hadn't run `go install` first, as we wouldn't fix up `$PATH`.

I think Roger's idea is better. It's slightly more verbose, but:
1) It's consistent everywhere - tests and standalone reproducers.
2) It keeps `exec` explicit, which is a good thing given that it's special. It uses `stdin`, sets `stdout` and `stderr`, runs as a separate process, can be run in the background, etc.

mvdan:
(see commit message)

Fixes #163.

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -24,7 +24,6 @@ import (
 // Keep list and the implementations below sorted by name.
 //
 // NOTE: If you make changes here, update doc.go.
-//
 var scriptCmds = map[string]func(*TestScript, bool, []string){
 	"cd":       (*TestScript).cmdCd,
 	"chmod":    (*TestScript).cmdChmod,
--- a/testscript/exe.go
+++ b/testscript/exe.go
@@ -42,11 +42,13 @@ func IgnoreMissedCoverage() {
 // code to pass to os.Exit. It's OK for a command function to
 // exit itself, but this may result in loss of coverage information.
 //
-// When Run is called, these commands will be available as
-// testscript commands; note that these commands behave like
-// commands run with the "exec" command: they set stdout
-// and stderr, and can be run in the background by passing "&"
-// as a final argument.
+// When Run is called, these commands are installed as regular commands in the shell
+// path, so can be invoked with "exec" or via any other command (for example a shell script).
+//
+// For backwards compatibility, the commands declared in the map can be run
+// without "exec" - that is, "foo" will behave like "exec foo".
+// This can be disabled with Params.RequireExplicitExec to keep consistency
+// across test scripts, and to keep separate process executions explicit.
 //
 // This function returns an exit code to pass to os.Exit, after calling m.Run.
 func RunMain(m TestingM, commands map[string]func() int) (exitCode int) {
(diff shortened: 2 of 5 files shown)


testscript/cmd.go
```diff
@@ -48,6 +48,15 @@ var scriptCmds = map[string]func(*TestScript, bool, []string){
 	"wait":     (*TestScript).cmdWait,
 }
 
+var builtinCmds map[string]bool
```
rogpeppe, line 51: Rather than doing this, how about just putting the check into the actual command function itself? That way it'll work even if the command overrides a builtin command too.

In `exe.go`:
```
			scriptCmds[name] = func(ts *TestScript, neg bool, args []string) {
				if ts.params.RequireExplicitExec {
					ts.Fatalf("use 'exec %s' rather than '%s' (because RequireExplicitExec is enabled)", args[0], args[0])
				}
				ts.cmdExec(neg, append([]string{name}, args...))
			}
```

(also above, a mild suggestion for a phrasing change to the error message because i'm not overly keen on "as per")

testscript/exe.go
```diff
@@ -42,11 +42,15 @@ func IgnoreMissedCoverage() {
 // code to pass to os.Exit. It's OK for a command function to
 // exit itself, but this may result in loss of coverage information.
 //
-// When Run is called, these commands will be available as
-// testscript commands; note that these commands behave like
-// commands run with the "exec" command: they set stdout
-// and stderr, and can be run in the background by passing "&"
-// as a final argument.
+// When Run is called, these commands will work when run via "exec".
```
rogpeppe, line 45: I think it should explicitly say that the commands are installed in the path and are hence available as normal commands (not just via `exec`) because that might not be entirely obvious to the casual reader.

How about something like this?

```
When Run is called, these commands are installed as regular commands in the shell
path, so can be invoked with "exec" or via any other command (for example a shell script).
```

rogpeppe (commented):
Thanks for doing this! LGTM in general modulo a few comments and suggestions.

testscript/cmd.go
mvdan: Much cleaner, thank you!

I took this route because, in my head, builtins could not be shadowed or overriden. That's a question for a different thread, though.

testscript/exe.go
mvdan: Good point, that's less ambiguous.

mvdan:
@rogpeppe ready for a final approval, I think.

rogpeppe (approved):
LGTM thanks!

PR #168, opened 2022-07-24
testscript: port "mv" and "! cmp" from upstream

### LINKED ISSUE #153: Port 'cmd/go: add mv and support "! cmp" in script tests'
From https://go-review.googlesource.com/c/go/+/380916

mvdan:
(see commit message)

Fixes #153.

### DIFF
--- a/testscript/doc.go
+++ b/testscript/doc.go
@@ -100,16 +100,16 @@ commands support this prefix. They are indicated below by [!] in the synopsis.
 The command prefix [cond] indicates that the command on the rest of the line
 should only run when the condition is satisfied. The predefined conditions are:
 
- - [short] for testing.Short()
- - [net] for whether the external network can be used
- - [link] for whether the OS has hard link support
- - [symlink] for whether the OS has symbolic link support
- - [exec:prog] for whether prog is available for execution (found by exec.LookPath)
- - [gc] for whether Go was built with gc
- - [gccgo] for whether Go was built with gccgo
- - [go1.x] for whether the Go version is 1.x or later
- - [unix] for whether the OS is Unix-like (that is, would match the 'unix' build
-   constraint)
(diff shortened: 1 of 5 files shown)


rogpeppe (changes requested):
LGTM, thanks, except that the new functionality needs to be added to the docs.

mvdan:
PTAL. The way gopls ran 1.19's gofmt, it shifted quite a bit of whitespace in doc.go, which I think we want anyway. To ease your review, ignore whitespace changes: https://github.com/rogpeppe/go-internal/pull/168/files?w=1

testscript/doc.go
```diff
@@ -122,102 +122,107 @@ Additional conditions can be added by passing a function to Params.Condition.
...
+  - mkdir path...
+    Create the listed directories, if they do not already exists.
+
+  - unquote file...
+    Rewrite each file by replacing any leading ">" characters from
+    each line. This enables a file to contain substrings that look like
+    txtar file markers.
+    See also https://godoc.org/github.com/rogpeppe/go-internal/txtar#Unquote
+
+  - mv path1 path2
```
rogpeppe, line 189: Looks like both unquote and mv are out of alphabetical order here (yeah, unquote is an old problem, probably my fault, but worth fixing now too, i reckon).

rogpeppe (changes requested):
LGTM modulo one last trivial thing.

testscript/doc.go
mvdan: hadn't noticed we were trying to keep them sorted :) fixed in https://github.com/rogpeppe/go-internal/compare/24c64e725dba574088acf1dc892bf5fae970160b..2310f38754e627b9db74eba3c5327b73666b4576.

rogpeppe (approved):
LGTM thanks!

PR #175, opened 2022-08-04
Test custom conditions

datacharmer:
Add a test for custom conditions and a brief example in the documentation.

### DIFF
--- a/testscript/doc.go
+++ b/testscript/doc.go
@@ -120,6 +120,23 @@ when testing.Short() is false.
 
 Additional conditions can be added by passing a function to Params.Condition.
 
+A simple example:
+
+		Condition: func(cond string) (bool, error) {
+			switch cond {
+			case "always_true":
+				return true, nil
+			case "always_false":
+				return false, nil
+			default:
+				return false, fmt.Errorf("unrecognized condition %s", cond)
+			}
+		},
+
+For more complex conditions that require parameters (such as [exec:program_name]) you can
+split the `cond` argument and evaluate the components. The function `TestScripts` in testscript_test.go
+shows such an approach.
+
 The predefined commands are:
 
   - cd dir
--- a/testscript/testdata/custom_conditions.txt
+++ b/testscript/testdata/custom_conditions.txt
@@ -1,0 +1,67 @@
+[!exec:echo] skip
+
+env upper_word=ABCD
+env lower_word=abcd
+
+[always_true] exec echo 'this is true'
+stdout 'this is true'
+
+exec echo ''
+
+[!always_true] exec echo 'this is true'
+! stdout .
+
+exec echo ''
+
+[always_false] exec echo 'this is false'
+! stdout .
+
+exec echo ''
+
+[!always_false] exec echo 'this is false'
+stdout 'this is false'
+
+exec echo ''
+
+[is_upper:ABCD] exec echo 'it is upper'
+stdout 'it is upper'
+
+exec echo ''
+
+[is_upper:$upper_word] exec echo 'it is again upper'
+stdout 'it is again upper'
+
+exec echo ''
+
+[is_upper:abcd] exec echo 'it is upper'
+! stdout .
+
+exec echo ''
+
+[is_upper:$lower_word] exec echo 'it is lower'
+! stdout .
+
+exec echo ''
+
+[!is_upper:ABCD] exec echo 'it is upper'
+! stdout .
+
+exec echo ''
+
+[is_lower:abcd] exec echo  'it is lower'
+stdout 'it is lower'
+
+exec echo ''
+
+[is_lower:$lower_word] exec echo  'it is again lower'
+stdout 'it is again lower'
+
+exec echo ''
+
+[is_lower:ABCD] exec echo 'it is lower'
+! stdout .
+
+exec echo ''
+
+[!is_lower:$lower_word] exec echo 'it is lower'
+! stdout .
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
-module github.com/datacharmer/go-internal
+module github.com/rogpeppe/go-internal
 
-go 1.18
+go 1.17
+
+require github.com/pkg/diff v0.0.0-20210226163009-20ebb0f2a09e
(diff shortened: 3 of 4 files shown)


testscript/doc.go
```diff
@@ -120,6 +120,35 @@ when testing.Short() is false.
...
+			// Assume condition name and args are separated by colon (":")
+			args := strings.Split(cond, ":")
+			// Empty condition is already managed in testscript.run()
+			name := args[0]
+			switch name {
+			case "is_upper":
+				if len(args) < 2 {
+					return false, fmt.Errorf("syntax: [is_upper:word]")
+				}
+				return strings.ToUpper(args[1]) == args[1], nil
```
mvdan, line 135: to make the example with an argument a bit more realistic, how about a variant of `exists`? Like the built-in command, but being useful as a condition, like `[exists:foo/bar] mycmd foo/bar`.

```
case "exists":
  _, err := os.Lstat(args[1])
  return err == nil, nil
```

testscript/doc.go
```diff
@@ -120,6 +120,35 @@ when testing.Short() is false.
...
+			case "always_true":
+				return true, nil
```
mvdan, line 137: similar to my other comment, I'd like this to be slightly more realistic: in practice I don't think anyone would write an "always true" condition :)

How about one that's tied to an env var? For example, I've used "slow" or "long" env vars to tell `go test` to do more extensive testing, kinda like the opposite of `go test -short`. So you could have a `[long]` that checks `os.Getenv("TEST_LONG") == "true"`, for instance.

testscript/doc.go
```diff
@@ -120,6 +120,35 @@ when testing.Short() is false.
 
 Additional conditions can be added by passing a function to Params.Condition.
 
+An example:
```
mvdan, line 123: No need for this to be a separate paragraph, IMO. You can attach it to the previous paragraph.

testscript/doc.go
```diff
@@ -120,6 +120,35 @@ when testing.Short() is false.
...
+
+		Condition: func(cond string) (bool, error) {
+			// Assume condition name and args are separated by colon (":")
```
rogpeppe, line 126: I think that parsing the condition here is somewhat overkill tbh, and makes the example a bit longer than it needs to be. I'm pretty sure that people writing conditions are up to the task of extrapolating from something a bit simpler.

How about just something like:

```
Condition: func(cond string) (bool, error) {
    // Note: conditions can be more complex than this (for example [exists:/some/file]).
    if cond != "someCondition" {
        return false, fmt.Errorf("unknown condition")
    }
    // someCondition is a place-holder for any conditional logic you might want.
    return someCondition(), nil
}
```

?

testscript/doc.go
```diff
@@ -120,6 +120,35 @@ when testing.Short() is false.
 
 Additional conditions can be added by passing a function to Params.Condition.
```
rogpeppe, line 121: Maybe worth mentioning this:

```
The function will only be called when all built-in conditions have been checked for.
```

testscript/testdata/custom_conditions.txt
```diff
@@ -0,0 +1,67 @@
+[!exec:echo] skip
```
rogpeppe, line 1: It would be nice to have a test that for the custom condition function returning an error, but that might be a bit awkward to do.

rogpeppe (commented):
Thanks for this! I think that the example might be a bit larger than needed (people rarely need custom conditions, so it seems a bit bulky as proposed), but other opinions might vary. WDYT?

testscript/testdata/custom_conditions.txt
datacharmer: I couldn't find a way to do that.

testscript/doc.go
datacharmer: I can do that, if everyone agrees. However, I wanted the example to include some text manipulation logic, because I was baffled when I first tried to define a condition. What was unclear to me was that custom commands are defined as a map of functions, while custom conditions are just one function for all.
I think the current example will save users some grief.

testscript/doc.go
datacharmer: Done

testscript/doc.go
datacharmer: removed

testscript/doc.go
datacharmer: Changed the example

testscript/testdata/custom_conditions.txt
datacharmer: UPDATE: I found a way to test the errors (see `testdata/custom_conditions_errors.txt`)

testscript/doc.go
bitfield: For what it's worth, I also feel this is likely to confuse people. But perhaps what this is really telling us is that the API should _be_ a map of condition names to `func(string) (bool, error)`. What do you think?

rudifa:
I found this PR while looking for discussions about custom conditions in testscript.

I could not find in this PR any doc that would explain its purpose and use cases; reading the proposed code changes lead me to think that its purpose is different from mine.

What I would like have is: execute a test step if an envvar evaluates to true and skip the step
silently otherwise, something like this

[iftrue DOVET] exec cue vet sample.cue
[!iftrue DOCMP] cmp sample.cue sample.got.cue

or simply

[DOVET] exec cue vet sample.cue
[DOCMP] cmp sample.cue sample.got.cue

My use case would be similar to [this](https://github.com/rudifa/cue-issues-fmt-comments/blob/main/testdata/2567-all-ast.txtar)
where I'd want to prefix some command with an envvar condition

```
[DOPRINT] exec cue-ast-print 2567.cue
[DOPRINT] cmp stdout 2567.ast
exec cue fmt 2567.cue
[DOCMP] cmp 2567.cue 2567.got.cue
[DOVET] exec cue vet 2567.cue

[DOPRINT] exec cue-ast-print 2567.0.cue
[DOPRINT] cmp stdout 2567.0.ast
exec cue fmt 2567.0.cue
[DOCMP] cmp 2567.0.cue 2567.0.got.cue
[DOVET] exec cue vet 2567.0.cue

# many more like this...
```

This would let me run the same script at different times with different purposes, while having  test cases with the related data (in, got, ast, ...) all in one script.

@rogpeppe @mvdan @datacharmer @bitfield What do you think?

PR #177, opened 2022-08-17
testscript: remove leading newline before results

bitfield:
Unlike `go test`, the `testscript` command prints a blank line before either 'PASS' or 'FAIL' and its other output:

```
testscript test.txtar

PASS
```

This makes it awkward to use in a loop over many scripts, one at a time. For example:

```sh
for TS in *.txtar; do
    echo -n "${TS}... "
    testscript ${TS}
done
```

The desired output is:

        test1.txtar... PASS
        test2.txtar... PASS
        ...

But instead, with this leading newline, we see:

        test1.txtar...
        PASS
        test2.txtar...
        PASS
        ...
        
This commit removes the newline.

### DIFF
--- a/testscript/testdata/big_diff.txt
+++ b/testscript/testdata/big_diff.txt
@@ -7,7 +7,6 @@ env
 cmpenv stdout stdout.golden
 
 -- stdout.golden --
-
 > cmp a b
 FAIL: $$WORK${/}dir${/}script.txt:1: large files a and b differ
 -- dir/script.txt --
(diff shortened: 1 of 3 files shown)


mvdan (approved):
I imagine the reason was to prevent joining PASS or FAIL with the previous line, i.e. without a newline separating them.

But we don't have any tests that break due to this change, and the only two that change do so in a very reasonable way, so... SGTM :)

PR #179, opened 2022-08-20
testscript: trim '.txtar' extension from subtest name

bitfield:
I noticed that, while the `.txt` extension is stripped from the script filename when constructing the subtest name passed to `t.Run`, the same is not done for scripts with a `.txtar` extension. This commit strips `.txtar` too.

Rather than add an explicit test for this, I changed the name of one of the existing test scripts from `.txt` to `.txtar`, so that the tests would fail unless the extension is correctly stripped.


mvdan (approved):
Nicely spotted; I must have missed this when I added support for txtar filename extensions.

PR #180, opened 2022-08-20
Use 0o prefix for octal literals, and `gofumpt` all files

bitfield:
Since Go 1.13, octal literals can be written using an `0o` prefix, instead of the (slightly) ambiguous `0`. Indeed, `gofumpt` insists on this, and those of us who use `gopls` with `gofumpt` enabled have our octal literals rewritten for us automatically.

This commit applies that change, along with some other minor `gofumpt`-age, including adding or removing blank lines and reformatting doc comments.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -46,7 +46,6 @@ type Cache struct {
 // to share a cache directory (for example, if the directory were stored
 // in a network file system). File locking is notoriously unreliable in
 // network file systems and may not suffice to protect the cache.
-//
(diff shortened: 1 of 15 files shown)


mvdan (approved):
Goes without saying that I like these changes, but perhaps that isn't a proper approval :wink:

mvdan:
This feels low risk, so I'll go ahead and merge. @rogpeppe can always send a follow-up PR or a revert if he disagrees.

PR #185, opened 2022-09-27
testscript: Fail testscript if txtar contains duplicate filenames

### LINKED ISSUE #184: testscript: Add check for duplicate and out-of-order files
I was just reviewing my testscripts and noticed that one of them had a duplicate file:

```
# test ...
exec command

-- filename --
# contents of filename
-- some/other/file --
# contents of some/other/file
-- filename --
# different contents of filename --
```

testscript did not warn about the duplication, and the duplication was not easy to spot as the files were out of order. If the files had been in order the duplicate files would have been easier to spot. Long story short, I ended up writing a very simple txtar linter and formater in https://github.com/twpayne/chezmoi/pull/2387.

Rather than adding a formatter and linter to testscript, I propose to add a mechanism similar to #164 and add two extra parameters:

`RequireUniqueFilenames bool`, default `false`, which will return an error if there are any duplicate filenames in the txtar archive.

`RequiredSortedFilenames bool`, default `false`, which will return an error if the files in the txtar archive are not in order.

This should maintain backward compatibility while also effectively adding optional opt-in linting capabilities.

Is this of interest? Would you accept a PR that implements this?

twpayne:
Fixes #184.


twpayne:
The test failure seems unrelated to this PR.

testscript/testdata/testscript_duplicate_name.txt
```diff
@@ -0,0 +1,34 @@
...
+testscript scripts-case-sensitive
+testscript scripts-normalized
+
+[darwin] ! testscript -unique-names scripts-case-insensitive
+[darwin] stdout 'duplicate name'
+
+[linux] testscript -unique-names scripts-case-insensitive
+
+[windows] ! testscript -unique-names scripts-case-insensitive
+[windows] stdout 'duplicate name'
```
mvdan, line 18: Do Mac and Windows _always_ use case insensitive filesystems?

mvdan (approved):
Thanks! Apologies it took so long to get back to you. There's also a conflict, but I imagine it's a trivial one.

testscript/testdata/testscript_duplicate_name.txt
twpayne: We both know that they don't *always* :) It's rare, but possible, for macOS users to run their systems with case sensitive filesystems, for example, even though this tends to break a large number of macOS applications.

The effort of making the tests detect whether the filesystem they're running on is case sensitive or not is fairly large. I think you'd have to create a file with a long random lowercase name (to avoid conflicts with any existing files) and test if it exists with the name uppercased. You'd have to be careful to use the exact directory that the tests run in. You can't use a temporary directory as it by be mounted as a different filesystem.

So, for now, I've removed these parts of the tests.

twpayne:
> There's also a conflict, but I imagine it's a trivial one.

The conflict is indeed trivial, but the tests no longer pass. Specifically, the test that testscript should return an error when duplicate names are encountered fails. I've traced this as far as verifying that `ts.Fatalf` is called in `ts.setup` when duplicate names are found, but this does **not** result in testscript returning an error. It should.

I suspect that #192 might be related. Specifically, does #192 cope with `ts.Fatalf` being called in `ts.setup`? I get the impression that the error is now silently swallowed somewhere rather than being returned. cc @rogpeppe

rogpeppe:
@twpayne You're right: there was indeed an issue in that area that I'd overlooked. I've proposed #193 to fix that.

mvdan:
Please rebase and try again :)

twpayne:
Thanks both! I've rebased and now the test passes :)

rogpeppe (commented):
LGTM with one suggestion, thanks!

testscript/testscript.go
```diff
@@ -431,6 +436,16 @@ func (ts *TestScript) setup() string {
 	ts.archive = a
 	for _, f := range a.Files {
 		name := ts.MkAbs(ts.expand(f.Name))
+		if ts.params.RequireUniqueNames {
+			switch _, err := os.Lstat(name); {
+			case err == nil:
+				ts.Fatalf("'%s' would overwrite '%s' (because RequireUniqueNames is enabled)", f.Name, name)
```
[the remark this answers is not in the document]
mvdan: Good point Roger, I missed this!

mvdan (approved):

PR #189, opened 2022-11-23
cmd/testscript: add a -continue flag to not stop at the first error

mvdan:
(see commit message)

### DIFF
--- a/cmd/testscript/testdata/continue.txt
+++ b/cmd/testscript/testdata/continue.txt
@@ -1,0 +1,16 @@
+# should support -continue
+unquote file.txt
+
+# Running with continue, the testscript command itself
+# should fail, but we should see the results of executing
+# both commands.
+! testscript -continue file.txt
+stdout 'grep banana in'
+stdout 'no match for `banana` found in in'
+stdout 'grep apple in'
+
+-- file.txt --
+>grep banana in
+>grep apple in
+>-- in --
+>apple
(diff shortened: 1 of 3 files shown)


cmd/testscript/testdata/continue.txt
```diff
@@ -0,0 +1,14 @@
+# should support -continue
+unquote file.txt
+
+# Running with continue, the testscript command itself
+# should fail, but we should see the results of executing
+# both commands.
+! testscript -continue file.txt
+stdout 'exec false'
```
rogpeppe, line 8: Alternatively, rather than relying on the presence of the `false` and `true` executables, how about just defining an `error` command that prints an error and returns a non-zero status. Then you avoid that dependency and also get to check that the failure output contains the output from failing commands.

rogpeppe: In fact, no need to define a new command - maybe just run a `grep` command that's expected to fail instead?

rogpeppe (approved):
LGTM modulo one suggestion for a slight test improvement, thanks (and good catch on `runT`!)

cmd/testscript/testdata/continue.txt
mvdan: nit: would also be nice to run the test script without continue, to double check that the behavior does change with the flag on.

mvdan:
I did forget a test, didn't I :)

cmd/testscript/testdata/continue.txt
myitcv: That's a really good point re `false` and `true`. I'll fix that up.

myitcv: Fixed.

mvdan:
Thanks! Will merge on green.

myitcv (approved):
LGTM. But I'm biased, because I helped write the test :)

PR #192, opened 2023-01-06
testscript: add Params.ContinueOnError

rogpeppe:
This fixes a few issues around the existing `testscript -continue` flag. Notably:

- causing `T.Fatal` and `T.FailNow` to return normally meant that some logic inside the testscript package that was expecting the old behaviour would fail to work correctly (for example, a non-existent command would cause a panic)
- the logging logic was erroneously assuming that if we got to the end of a script, all sections had passed therefore we could omit the logged messages.

We fix the above by making "continue on error" a first class part of `testscript.Params`, so the testscript logic actually knows about it.

To test it properly, we need a couple of hacks to make the output predictable. Unfortunately, those hacks are internal only, so it's hard to add similar tests to the `cmd/testscript` command to end-to-end test that, but the existing test should act as sufficient "smoke test" that the continue logic is wired up OK.

### DIFF
--- a/cmd/testscript/main.go
+++ b/cmd/testscript/main.go
@@ -209,8 +209,9 @@ func (tr *testRunner) run(runDir, filename string) error {
 	}
 
 	p := testscript.Params{
-		Dir:           runDir,
-		UpdateScripts: tr.update,
+		Dir:             runDir,
+		UpdateScripts:   tr.update,
+		ContinueOnError: tr.continueOnError,
 	}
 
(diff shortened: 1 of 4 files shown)


mvdan (approved):
globals modified by tests always make me worry a bit - I wonder if we could use something other than `cmpenv` here, like having a glob or regular expression as a placeholder for elapsed times and env var lines.

not super worried about it, though, and this test is certainly better than what I had done before.

PR #194, opened 2023-01-18
ci: bump Go versions

twpayne:
I noticed CI was still using Go 1.17 and 1.18.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,23 +11,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.17.x, 1.18.x]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        go-version:
+        - '1.18.x'
+        - '1.19.x'
+        os:
+        - ubuntu-latest
+        - macos-latest
+        - windows-latest
     runs-on: ${{ matrix.os }}
     steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
     - name: Install Go
       uses: actions/setup-go@v3
       with:
         go-version: ${{ matrix.go-version }}
-    - name: Checkout code
-      uses: actions/checkout@v3
     - name: Test
       run: |
         go test ./...
         go test -race ./...
 
     - name: Tidy
-      if: matrix.os == 'ubuntu-latest' # no need to do this everywhere
+      if: matrix.os == 'ubuntu-latest' && matrix.go-version == '1.19.x' # no need to do this everywhere
       run: |
         go mod tidy
 
--- a/cmd/txtar-c/savedir.go
+++ b/cmd/txtar-c/savedir.go
@@ -10,7 +10,6 @@
 //
 // See https://godoc.org/github.com/rogpeppe/go-internal/txtar for details of the format
 // and how to parse a txtar file.
-//
 package main
 
 import (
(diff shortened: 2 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -40,15 +45,15 @@ jobs:
...
+        GO_COMMIT: 6fcbd0369d3ac9002e2b46229f90ce64ce0539e4 # 2023-01-18
       run: |
         cd $HOME
         mkdir $HOME/gotip
         cd $HOME/gotip
 
         wget -O gotip.tar.gz https://go.googlesource.com/go/+archive/${GO_COMMIT}.tar.gz
         tar -xf gotip.tar.gz
-        echo "devel go1.19-${GO_COMMIT}" >VERSION
+        echo "devel go1.20-${GO_COMMIT}" >VERSION
```
mvdan, line 56: note that master is 1.21 now :)

.github/workflows/test.yml
```diff
@@ -11,23 +11,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.17.x, 1.18.x]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        go-version:
+        - stable
+        - oldstable
```
mvdan, line 16: I'm personally not a fan of this, because in my opinion we want to be conscious about what major versions of Go we support.

We do tend to do the last two versions, but when a new version comes out, it's a human process to update the versions in multiple places (CI, go.mod, etc) and test that everything still works. Sometimes a few tests break, especially since we have tight coupling with some packages like testing - and we also don't want CI to start failing when a new version of Go is out :)

but perhaps @rogpeppe disagrees. Some people prefer their CI to track "latest stable" with the Go version, and then they can react when CI starts failing on its own. I personally prefer to keep CI closer to determinism, so that it's not likely to start failing on its own.

mvdan (commented):
Thanks!

.github/workflows/test.yml
mvdan: Worth noting that depending on the project I also track versions of the operating systems manually. For example, on a project of mine, `macos-latest` broke a test when that jumped from `macos-11` to `macos-12`. But, for most Go projects, OS upgrades don't change behavior.

.github/workflows/test.yml
twpayne: Good point! In any case I've reverted this change as the tests fail with the latest commit from `master`.

.github/workflows/test.yml
twpayne: Ack, I've changed the PR to use Go versions `1.18.x` and `1.19.x` explicitly.

.github/workflows/test.yml
```diff
@@ -11,23 +11,28 @@ jobs:
...
-    - name: Checkout code
-      uses: actions/checkout@v3
     - name: Test
       run: |
         go test ./...
         go test -race ./...
 
     - name: Tidy
-      if: matrix.os == 'ubuntu-latest' # no need to do this everywhere
+      if: matrix.os == 'ubuntu-latest' && matrix.go-version == 'stable' # no need to do this everywhere
```
mvdan, line 35: presumably this is wrong as it still mentions "stable"?

mvdan (approved):

.github/workflows/test.yml
twpayne: /facepalm Good catch, thanks.

mvdan (approved):

PR #198, opened 2023-02-03
testscript: suggest misspelled commands

### LINKED ISSUE #190: testscript: Accept !cmd (without space) or provide better error
A colleague is trying out `testscript` and used `!cmd` to call a custom command (note the lack of space). They got the error message "`!cmd` not defined" and where confused why the negation was not recognized. I'd like to improve that.

One option would be to accept the `!cmd`, parsing it as equivalent to `! cmd`. This is easy to implement, but probably not backwards compatible.

Another option is to provide a better error message if a command is not found. e.g. if `!cmd` is looked up and not found, but `cmd` is, we could detect that and report "did you mean `! cmd`?". Or - harder to implement, but more generally useful - we could search the defined commands by edit-distance to the missed command and if one of them is a close enough match, suggest that. [x/tools has an internal package that might help](https://pkg.go.dev/golang.org/x/tools/internal/lsp/fuzzy).

I can send a PR, but wanted to discuss what option you prefer.

Merovius:
If a command is not found, we go through the list of defined commands and check if any of them are sufficiently close to the one used. "Sufficiently close" is defined by having a Damerau-Levenshtein distance of 1, which feels like it hits the sweet spot between usefulness and ease of implementation.

The negation case is still special-cased, as negation is not in the set of defined commands.

Fixes #190


Merovius:
Hm I just realized that this might create undesirable results for the `!cmd` case, as it will result in both suggesting `! cmd` *and* `cmd`. I probably should special-case this even further (which means all the spelling-correction stuff is useless for actually fixing #190 specifically. Still useful, though).

Merovius:
Okay, second try :) Also over engineered the test a bit and actually found what's arguably a bug in handling invalid UTF-8 by better fuzzing.

Merovius:
@mvdan You offered to review fixes to #190. FWIW I'd totally accept being told I'm overengineering things and re-trying with something simpler.

ldemailly:
I propose #215 for the original problem :-) (and I'm ok if that doesn't get merged, though I think it's simple/straightforward, I'll maintain on my fork meanwhile)

internal/textutil/almost_equal.go
```diff
@@ -0,0 +1,67 @@
+package textutil
```
mvdan, line 1: can we come up with a better package name? :) how about `textdistance` or `misspell`?

testscript/testscript.go
```diff
@@ -661,12 +663,56 @@ func (ts *TestScript) runLine(line string) (runOK bool) {
...
+	}
+	for c := range ts.params.Cmds {
+		if textutil.AlmostEqual(name, c) {
+			candidates = append(candidates, c)
+		}
+	}
+	if len(candidates) == 0 {
+		return nil
+	}
+	// deduplicate candidates
```
mvdan, line 705: add a TODO to use `slices.Sort` plus `slices.Compact` once those hit std; in the meantime, I think relying on x/exp would be a bit too risky for a library.

mvdan (approved):
This is quite a bit of added code, but it's useful and you've clearly put effort and polish into it, and I don't think it will cause additional maintenance burden of its own. SGTM with a couple of nits. There are conflicts as well, but I imagine they will be trivial.

internal/textutil/almost_equal.go
Merovius: The `textutil` is pre-existing, only containing `Diff`. It felt like a logical place to put this. I'm fine putting it in a package on its own, just making sure you didn't miss that.

testscript/testscript.go
Merovius: Done, though `slices.Sort` is pending golang/go#59488 FWIW

internal/textutil/almost_equal.go
mvdan: Huh, I totally missed that. The last use of textutil went away when we made packages like `modfile` simply forward to `x/mod`, so the package is now unused and IMO should be deleted. Please pretend it's not there; I will delete it soon.

Merovius:
I feel like there should be a test that the code in `testscript` actually does what it's supposed to. I'm trying to figure out how the testscript tests work to make that happen.

mvdan:
Look at files like `testscript/testdata/testscript_update_script.txt`; we have "meta-testscripts" that themselves set up a testscript and run it, sometimes with a flag, sometimes expecting a failure :) The extra level allows this kind of flexibility.

Merovius:
Added a couple simple tests and renamed the package PTAL

testscript/testdata/testscript_notfound.txt
```diff
@@ -0,0 +1,9 @@
+# Check that unknown commands output the right error
```
mvdan, line 1: what about joining all three scripts into one, with sections for each of the three cases? a total of ~30 lines doesn't feel like enough to warrant splitting into multiple files. you can name the inner scripts like `-- scripts/notfound_misspelled.txt --`.

mvdan (approved):
Thanks! Just one last nit about the tests.

testscript/testdata/testscript_notfound.txt
Merovius: I created multiple files because I couldn't figure out how to tell it what output to expect for which script. Not terribly fluent in `testscript` :) How about this?

mvdan: SGTM!

PR #203, opened 2023-02-15
robustio: add package

eandre:
This adds the `robustio` package from [golang.org/x/tools/internal/robustio](https://pkg.go.dev/golang.org/x/tools/internal/robustio) which in turn is copied from `cmd/go/internal/robustio`.

It would be great to add this to go-internal as we've sort of been standardizing on this repo for a lot of go-internal utility packages. I could of course maintain this myself but I hope you consider it for inclusion. The errors it "fixes" (papers over) are quite common when dealing with temporary files on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Included are the following:
 - modfile: read and write `go.mod` files while preserving formatting and comments.
 - module: module paths and versions.
 - par: do work in parallel.
+- robustio: transparent retrying of failure-prone I/O functions on Windows.
 - semver: semantic version parsing.
 - testenv: information on the current testing environment.
 - testscript: script-based testing based on txtar files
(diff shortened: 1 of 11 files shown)


mvdan:
Historically we've only copied packages from Go's main repository, not x/tools, so I would lean towards including https://pkg.go.dev/cmd/go/internal/robustio instead of https://pkg.go.dev/golang.org/x/tools/internal/robustio.

Also note that the x/tools API is slightly larger, with https://pkg.go.dev/golang.org/x/tools/internal/robustio#GetFileID. I would argue that not including that is a good thing, as it seems like an opaque and low level API that doesn't seem useful on its own.

mvdan (changes requested):
Marking as "changes requested" to remind myself :)

rogpeppe:
Superceded by #239 (thanks for the suggestion @eandre!)

PR #205, opened 2023-03-22
diff: add package, make testscript use it

### LINKED ISSUE #157: diff: add new diff package based on Go 1.19 patience diff implementation
Hi there👋, I would be interested in sending a PR to add a new diff package to go-internal based on the [patience diff](https://bramcohen.livejournal.com/73318.html?noscroll&utm_medium=endless_scroll) implementation Russ added in [CL 384255](https://golang.org/cl/384255/) during 1.19 dev cycle. 

Compared to the the current pkg/diff used by go-internal, it is faster and much more memory efficient, running in O(n log n) time and I think it is O(n + m) space.

I currently have a copy here:
   https://github.com/thepudds/patience-diff

I think go-internals would be a better home for it, including I think it is generally useful for the community, it would bring go-internal down to zero external dependencies (which is part of my selfish reason for being interested in a PR), as well as help deal w/ current testscript diff performance issues such as:

https://github.com/rogpeppe/go-internal/blob/master/testscript/cmd.go#L144-L148
```go
	// pkg/diff is quadratic at the moment.
	// If the product of the number of lines in the inputs is too large,
	// don't call pkg.Diff at all as it might take tons of memory or time.
	// We found one million to be reasonable for an average laptop.
	const maxLineDiff = 1_000_000
```	

A sample performance comparison (which I originally posted in https://github.com/pkg/diff/issues/26#issuecomment-1045170532):

> For this example [~130k lines], it runs in 0.07 sec and uses 38 MB RSS, vs. pkg/diff runs in 19 sec and 18 GB RSS (roughly ~250x slower and ~500x more RAM).
> [...]
> The patience-diff version is not a minimal diff:
> 
> ```
> $ wc -l *.out
> 96935 patience-diff.out
> 56391 pkg-diff.out
> ```
> ...but at least in theory, it is often friendlier to humans.

mvdan:

Fixes #157.


myitcv (approved):
LGTM - but would also ask @rogpeppe to confirm the change in package "API" surface.

mvdan:
CC @thepudds

thepudds:
Hi @mvdan , looks great, and thanks! (Sorry, this was still on my to do list, but not my "Burning Must Do Now" list 😅 ).

Would it make sense to remove or modify as well (currently still there on `master`):

https://github.com/rogpeppe/go-internal/blob/50a14409f175f086f5f0f343bf3e06ca917da08b/testscript/cmd.go#L144-L152

mvdan:
Oh my, how did I miss that? Happy to approve and merge a PR if you send one, otherwise I'll send one in a bit and you can approve.

PR #212, opened 2023-03-27
renameio: implement in terms of github.com/google/renameio/v2

rogpeppe:


mvdan (commented):
I'm just reminded that the public renameio does not support Windows by design; see https://pkg.go.dev/github.com/google/renameio/v2#readme-windows-support.

There is https://pkg.go.dev/github.com/google/renameio/v2@v2.0.0/maybe#WriteFile which is best-effort, falling back to `ioutil.WriteFile` on Windows. We might want to use that for `WriteFile`, to not break Windows, but then I'm not sure what we should do with `WriteToFile`.

Maybe we don't forward this package due to the Windows awkwardness, but only deprecate it?

rogpeppe:
> Maybe we don't forward this package due to the Windows awkwardness, but only deprecate it?

Yeah, not a bad idea. I'll abandon this PR.

PR #213, opened 2023-03-27
all: deprecate wrapper packages

rogpeppe:
People should in general be using the upstream packages rather than these. Also use the upstream packages directly when possible and remove some tests that aren't testing code in this module any more.


mvdan (approved):
Nice, this will also make the go-internal module zip noticeably smaller :)

PR #214, opened 2023-03-29
Fix tests

ldemailly:


ldemailly:
opsa, wrong against which to create mr my bad

PR #215, opened 2023-03-29
Accept `!` for negation without requiring a space

### LINKED ISSUE #190: testscript: Accept !cmd (without space) or provide better error

ldemailly:
Fixes #190

### DIFF
--- a/testscript/testdata/negate_no_space.txt
+++ b/testscript/testdata/negate_no_space.txt
@@ -1,0 +1,5 @@
+[!exec:false] skip
+
+!exec false
+!stderr .
+!stdout .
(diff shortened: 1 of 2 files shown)


ldemailly:
for the record, re consistency:


lines like

[!exec:false] skip

aren't

[ ! exec:false ] skip

PR #216, opened 2023-04-25
testscript: expose (*TestScript).stdout via Stdout()

myitcv:
(see commit message)

### DIFF
--- a/testscript/testscript.go
+++ b/testscript/testscript.go
@@ -14,6 +14,7 @@ import (
 	"flag"
 	"fmt"
 	"go/build"
+	"io"
 	"io/fs"
 	"io/ioutil"
 	"os"
@@ -362,6 +363,17 @@ type TestScript struct {
 	scriptFiles   map[string]string           // files stored in the txtar archive (absolute paths -> path in script)
 	scriptUpdates map[string]string           // updates to testscript files via UpdateScripts.
 
+	// runningBuiltin indicates if we are running a user-supplied builtin
+	// command. These commands are specified via Params.Cmds.
+	runningBuiltin bool
+
+	// builtinStd(out|err) are established if a user-supplied builtin command
+	// requests Stdout() or Stderr(). Either both are non-nil, or both are nil.
+	// This invariant is maintained by both setBuiltinStd() and
+	// clearBuiltinStd().
+	builtinStdout *strings.Builder
+	builtinStderr *strings.Builder
+
 	ctxt        context.Context // per TestScript context
 	gracePeriod time.Duration   // time between SIGQUIT and SIGKILL
 }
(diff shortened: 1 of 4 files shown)


myitcv:
Noting previous exchanges in https://github.com/rogpeppe/go-internal/pull/140

testscript/testscript.go
```diff
@@ -788,6 +789,27 @@ func (ts *TestScript) Check(err error) {
...
+func (ts *TestScript) Stderr() io.Writer {
+	return stringAppender{&ts.stderr}
+}
+
+type stringAppender struct {
+	s *string
+}
+
+func (sa stringAppender) Write(p []byte) (n int, err error) {
+	*sa.s = *sa.s + string(p)
```
mvdan, line 809: perhaps leave a TODO to use a strings.Builder at some point; this is going to be quadratic and pretty inefficient.

mvdan (approved):
I assume we're OK with Stdout/Stderr not being safe for concurrent use.

testscript/testscript.go
myitcv: I'm not clear that we can use `strings.Builder`. The underlying `stdout` field is a `string`, and there is a lot of usage of that field assuming it is a `string`. i.e. it would be a significant change in the semantics of the underlying field if we were to make that change.

Perhaps we should, therefore, expose `(*TestScript).SetStdout(string)` instead. Because it better reflects what's going on.

myitcv: Alternative approach in https://github.com/rogpeppe/go-internal/pull/217. Indeed I think I prefer that approach because it places the burden of "gathering" stdout on the caller.

mvdan: I slightly prefer io.Writer APIs, but I agree it goes against the internal "string value" design, so using the other PR seems fine to me. Another disadvantage of this API is that `TestScript.Stdout` could be understood as a stdout getter/reader rather than setter or writer, and `TestScript.SetStdout` doesn't have that issue.

myitcv:
@mvdan @rogpeppe - PTAL

testscript/testscript.go
```diff
@@ -362,6 +363,17 @@ type TestScript struct {
...
+	// runningBuiltin indicates if we are running a user-supplied builtin
+	// command. These commands are specified via Params.Cmds.
+	runningBuiltin bool
+
+	// builtinStd(out|err) are established if a user-supplied builtin command
+	// requests Stdout() or Stderr(). Either both are non-nil, or both are nil.
+	// This invariant is maintained by both setBuiltinStd() and
+	// clearBuiltinStd().
+	builtinStdout *bytes.Buffer
+	builtinStderr *bytes.Buffer
```
mvdan, line 375: nit: use strings.Builder, since all you do with the result is call the String method

testscript/testscript.go
```diff
@@ -788,6 +819,60 @@ func (ts *TestScript) Check(err error) {
...
+	return ts.builtinStdout
+}
+
+// Stderr returns an io.Writer that can be used by a user-supplied builtin
+// command (delcared via Params.Cmds) to write to stderr. If this method is
+// called outside of the execution of a user-supplied builtin command, the
+// call panics.
+func (ts *TestScript) Stderr() io.Writer {
+	if !ts.runningBuiltin {
+		panic("can only call Stderr() when running a builtin command")
```
mvdan, line 840: nit: refer to the methods like `TestScript.Stderr` and without parentheses, for consistency with godoc

mvdan (approved):
SGTM

testscript/testscript.go
myitcv: Done. I always forget this one...

testscript/testscript.go
myitcv: Done.

PR #217, opened 2023-04-26
testscript: expose (*TestScript).stdout via SetStdout(string)

myitcv:
(see commit message)

### DIFF
--- a/testscript/testdata/cmd_stdout_stderr.txt
+++ b/testscript/testdata/cmd_stdout_stderr.txt
@@ -1,0 +1,15 @@
+# Print to stdout via a TestMain cmd
+fprintargs stdout hello world from fprintargs
+stdout 'hello world from fprintargs'
+
+# Verify that we can write to stdout via a Params.Cmds cmd
+echoStdout hello world from echoStdout
+stdout 'hello world from echoStdout'
+
+# Print to stderr via a TestMain cmd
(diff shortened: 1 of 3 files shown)


mvdan (approved):
SGTM with two nits:

* Please add a test case for a command that writes to stdout and stderr, and then returns an error. I wonder if stderr is then kept at all, for example.
* The wording for "updates the last stdout/stderr" feels slightly confusing; it feels like I could/should call it multiple times, and "last" isn't entirely correct when one is running the current command. How about "SetStdout can be used by commands in Params.Cmds to provide a resulting stdout string", since I don't believe we have another usecase yet?

myitcv:
> * Please add a test case for a command that writes to stdout and stderr, and then returns an error. I wonder if stderr is then kept at all, for example.

I've replied to this point in https://github.com/rogpeppe/go-internal/issues/139#issuecomment-1522914966 so we keep decisions in the issue (because it feels like I'm straying from the PR being "just" the implementation here).

> * The wording for "updates the last stdout/stderr" feels slightly confusing; it feels like I could/should call it multiple times, and "last" isn't entirely correct when one is running the current command. How about "SetStdout can be used by commands in Params.Cmds to provide a resulting stdout string", since I don't believe we have another usecase yet?

Will fix this as/when we agree on the API/approach.

myitcv:
Closing in favour of #216

PR #218, opened 2023-04-29
update cache and lockedfile packages

mvdan:

Particularly the cache package, since I want to use it soon and it hadn't been updated since 2018.


mvdan:
Should be passing CI now; I had to use a slightly older version of `lockedfile` to prevent a very recent refactor that broke 1.19 and 1.20.

rogpeppe (approved):

PR #221, opened 2023-05-04
testscript: fix "signal: killed" exec errors on MacOS by using os.Symlink on darwin

bep:
By using `os.Symlink` on Darwin.

See #200

### DIFF
--- a/testscript/exe.go
+++ b/testscript/exe.go
@@ -123,7 +123,10 @@ func RunMain(m TestingM, commands map[string]func() int) (exitCode int) {
 // as that can lead to "access denied" errors when removing.
 func copyBinary(from, to string) error {
 	if runtime.GOOS != "windows" {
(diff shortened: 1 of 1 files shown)


mvdan:
I wonder if we should support hard links at all, then. I chose hard links because symlinks can cause more trouble, but empirically, hard links are causing trouble as well. Symlinks are better on MacOS, and they shouldn't be worse on Linux. 

If you replace `os.Link` with `os.Symlink`, and add a brief comment about hard links having caused issues on Mac before, I'm happy to merge - with the assumption that we're pretty sure the flakes on Mac would be gone.

bep:
@mvdan I have force pushed a version that uses `os.Symlink` for all (except Windows). I have also repeated my test a few times, and it 1. Never fails with this branch 2. Always fails on master.

PR #222, opened 2023-05-05
testscript: use unix.CloneFile on MacOs 

### LINKED ISSUE #200: testscript: "signal: killed" exec errors on MacOS 12

bep:
To fix unexpected errors of type:

```
[signal: killed]
FAIL: testscripts/myecho.txt:1: unexpected command failure
```

Fixes #200

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,6 @@ go 1.19
 
 require (
 	golang.org/x/mod v0.9.0
+	golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f
 	golang.org/x/tools v0.1.12
 )
--- a/go.sum
+++ b/go.sum
@@ -1,4 +1,6 @@
 golang.org/x/mod v0.9.0 h1:KENHtAZL2y3NLMYZeHY9DW8HW8V+kQyJsY/V9JlKvCs=
 golang.org/x/mod v0.9.0/go.mod h1:iBbtSCu2XBx23ZKBPSOrRkjjQPZFPuis4dIYUhu/chs=
+golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f h1:v4INt8xihDGvnrfjMDVXGxw9wrfxYyCjk0KbXjhR55s=
+golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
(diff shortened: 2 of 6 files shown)


mvdan:
Failed build, FYI - otherwise SGTM. cc @myitcv @rogpeppe

bep:
@mvdan build should be fixed now (build constrains can be tricky to get right ...)

mvdan (approved):
SGTM; merging relatively quickly since the issue is relatively severe for Mac users, and the fix being in master will allow for more testing.

Issue #196, opened 2023-01-21
move to golang or other 'official' org?


ldemailly:
Maybe more a question than an issue but why isn't this repo like

https://github.com/golang/ ..

or

https://github.com/google/go-cmdtest

ie under an org rather than an individual? I think `testscript` is awesome and it would make it easier to adopt and publicize under a. well known (or even its own) org ?

mvdan:
See https://github.com/go-quicktest/qt, which I understand @frankban and @rogpeppe want to switch to at some point. Moving to an org before then is likely not a good idea, as it would be a breaking change.

ldemailly:
Thanks for the quick response. So far qt doesn't seem to have txtar etc support but are you saying it'll be added there? (I didn't see it mentioned in the v1 issue either)

mvdan:
That new location would be for quicktest, which is what you asked for. Other parts of go-internal are now being exposed as libraries by upstream, like x/tools/txtar, so I think the rest of go-internal will be deprecated over time

ldemailly:
I am asking specifically about `testscript` ie a way to test go cli

mvdan:
Right, I got confused between testscript and quicktest. I don't think there are plans to move it to an org, but @rogpeppe can always clarify.

ldemailly:
thoughts?

bep:
I agree that this deserves some broader audience. I stumbled upon this "by accident" and txtar and testscript have been (in that order) the most time saving new "things" I have added to my tool box the last decade.

myitcv:
This is sort of happening but in a slightly different form to what you might imagine. The following packages already exist in a more "official" namespace:

```
golang.org/x/mod/modfile
golang.org/x/mod/module
golang.org/x/mod/semver
golang.org/x/tools/txtar
```

The `testscript` portion was carved out of `cmd/go` internals into https://github.com/golang/go/tree/master/src/cmd/go/internal/script. That is, granted, still internal. But at least moving to a more reusable state.

My main question, @ldemailly and @bep, is: what are you trying to solve for by moving this to a more official org? 

Because I'm not clear why its current location precludes such publicity, e.g.

https://bitfieldconsulting.com/golang/test-scripts

cc @bitfield for thoughts too

bep:
>My main question, @ldemailly and @bep, is: what are you trying to solve for by moving this to a more official org?

[...]

PR #225, opened 2023-05-07
readme: provide more context on maintainers and give a testscript overview with links

thepudds:
Help give some context for people who might be evaluating using this repo and are wondering things like:

* what is the bus factor for the repo?
* who are the maintainers?
* are the maintainer(s) of this repo still interested in this repo (including, are they still using it)?
* are other people using testscript?
* where did testscript come from?

Obviously, someone can hunt around to determine many of these things on their own, but intent is to help make that process more efficient.

Updates #196

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,12 @@
 This repository factors out an opinionated selection of internal packages and functionality from the Go standard
 library. Currently this consists mostly of packages and testing code from within the Go tool implementation.
 
+This repo is [primarily maintained](https://github.com/rogpeppe/go-internal/graphs/contributors) by long-time [Go contributors](https://github.com/golang/go/contributors) who are also currently [maintaining CUE](https://github.com/cue-lang/cue/graphs/contributors) (which is primarily written in Go and which relies upon several of the packages here).
+
+Contributions are welcome, but please open an issue for discussion first.
+
+### Packages
+
 Included are the following:
 
 - dirhash: calculate hashes over directory trees the same way that the Go tool does.
@@ -15,6 +21,16 @@ Included are the following:
 - testscript: script-based testing based on txtar files
 - txtar: simple text-based file archives for testing.
 
-# Links
+### testscript
 
-- [Test scripts in Go](https://bitfieldconsulting.com/golang/test-scripts)
+The most popular package here is the [testscript](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript) package:
+ * Provides a shell-like test environment that is very nicely tuned for testing Go CLI commands.
+ * Extracted version of the core Go team's internal testscript package.
+ * [Heavily used](https://github.com/golang/go/tree/master/src/cmd/go/testdata/script) to test the `go` command.
+ * Includes patterns for checking stderr/stdout, command pass/fail assertions, and so on.
+ * Built-in support for Go concepts like build tags.
+ * Integrates well with `go test`, including coverage support.
+ * Inputs and sample outputs are in the [txtar](https://pkg.go.dev/golang.org/x/tools/txtar) text archive format, also used by the Go playground.
+ * Accompanied by a [testscript](https://github.com/rogpeppe/go-internal/tree/master/cmd/testscript) command for running standalone scripts with files embedded in txtar format.
+ 
+ A nice introduction to using testscripts is this [blog post](https://bitfieldconsulting.com/golang/test-scripts) series.


bitfield:
It's worth knowing that the `Links` header [is special](https://github.com/golang/go/issues/42968): it causes any listed links to be shown in the right-hand pane on pkg.go.dev.

thepudds:
Hi @bitfield 👋 

> It's worth knowing that the Links header https://github.com/golang/go/issues/42968: it causes any listed links to be shown in the right-hand pane on pkg.go.dev.

Ah, interesting, TIL.

I guess I don't know how often people are browsing the root https://pkg.go.dev/github.com/rogpeppe/go-internal on pkg.go.dev (vs. browsing API for an importable package), and I guess the "Links" link in the right-hand pane hasn't been jumping out to me personally.

My thought was to put the link in a sentence along with words like "introduction" and "nice" and "series" to encourage interested people to click, but I don't have a strong opinion.

As the target of said link, any preference regarding which form to use?

ldemailly:
I tested with https://github.com/fortio/testscript/blob/main/README.md and you can use ## for links so it's smaller (and I guess probably ### or #### would also work)

README.md
```diff
@@ -15,6 +21,16 @@ Included are the following:
...
 - txtar: simple text-based file archives for testing.
 
-# Links
+### testscript
 
-- [Test scripts in Go](https://bitfieldconsulting.com/golang/test-scripts)
+The most popular package here is the [testscript](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript) package:
+ * Provides a shell-like test environment that is very nicely tuned for testing Go CLI commands.
+ * Extracted version of the core Go team's internal testscript package.
+ * [Heavily used](https://github.com/golang/go/tree/master/src/cmd/go/testdata/script) to test the `go` command.
```
myitcv, line 29: Having this comment as a separate bullet seems to suggest that `github.com/rogpeppe/go-internal/testscript` is used in the `cmd/go` tests, which it is not. I would perhaps fix that by connecting this bullet to the previous one, and adjusting the wording to "link" such that it is clear the internal "version" of `testscript` is used for `cmd/go` tests. Perhaps even linking to and naming https://github.com/golang/go/tree/master/src/cmd/go/internal/script.

README.md
```diff
@@ -1,6 +1,12 @@
 This repository factors out an opinionated selection of internal packages and functionality from the Go standard
 library. Currently this consists mostly of packages and testing code from within the Go tool implementation.
 
+This repo is [primarily maintained](https://github.com/rogpeppe/go-internal/graphs/contributors) by long-time [Go contributors](https://github.com/golang/go/contributors) who are also currently [maintaining CUE](https://github.com/cue-lang/cue/graphs/contributors) (which is primarily written in Go and which relies upon several of the packages here).
```
myitcv, line 4: I don't think we've mentioned this anywhere, but I tend to find reviewing long lines (of markdown) very hard.

Indeed we tend to follow https://sembr.org/ in the CUE project. 

With a rough guide of breaking at 100 characters unless there a natural semantic break before.

README.md
```diff
@@ -15,6 +21,16 @@ Included are the following:
 - testscript: script-based testing based on txtar files
 - txtar: simple text-based file archives for testing.
 
-# Links
+### testscript
```
myitcv, line 24: Did you intend for this to be a level 4 heading below `### packages`?

myitcv (approved):
Thanks very much, @thepudds.

LGTM with a couple of suggestions.

myitcv (approved):
Thanks very much, @thepudds.

LGTM with a couple of suggestions.

mvdan:
SGTM :) Perhaps worth making a mention that testscript and txtar were both originally written by Russ. We don't want to give the impression that we came up with the idea.

README.md
thepudds: Done.

README.md
thepudds: (A level 4 heading made testscript maybe too small, so went the other direction and made Packages a level 2 heading).

README.md
thepudds: Good point. 

(I originally wrote these bullets as a paragraph with more connective words, but then thought bullets would be better, and in the process created the problem you point out).

thepudds:
Thanks for the quick feedback!  I think I addressed your comments. 

PTAL.

README.md
myitcv: Great, thanks.

ldemailly:
You don't want to keep/add back ### Links for pkg.go.dev?

mvdan (approved):
This repo is under @rogpeppe's name, so I'd like a review from him as well.

rogpeppe (approved):
LGTM, thanks!

PR #247, opened 2024-02-19
testscript: Add Chdir method to change directory

abhinav:
It is not currently possible for a custom testscript command
to change the working directory of that script run.
`TestScript.Exec` runs the command in a subprocess,
so one cannot do `ts.Exec("cd", dir)`.

This change adds a `Chdir` method to `TestScript`
that allows changing the working directory of the script.
The implementation is the same as the "cd" command,
which now relies on `Chdir`.

The availability of this function matches similar functionality in the
[`State.Chdir` method of rsc.io/script][1].
(I ported some tests from rsc.io/script to testscript.)

  [1]: https://pkg.go.dev/rsc.io/script#State.Chdir

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -62,18 +62,11 @@ func (ts *TestScript) cmdCd(neg bool, args []string) {
 	}
 
 	dir := args[0]
-	if !filepath.IsAbs(dir) {
-		dir = filepath.Join(ts.cd, dir)
-	}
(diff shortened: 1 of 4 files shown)


abhinav:
Gentle bump, @mvdan @rogpeppe.
Any interest in this PR?

mvdan (approved):
Thank you!

PR #248, opened 2024-02-19
all: Move away from ioutil

abhinav:
It's been a while since io/ioutil has been deprecated.
Go 1.22 was just released, and the go.mod specifies 'go 1.20'.
This should be safe to remove at this point.

The change was largely automated, minus the `t.TempDir()` calls inside
test files, which was done manually.


mvdan (approved):
Thanks!

Issue #234, opened 2023-08-08
Command go allowed in some txtar scripts and not in others, please document


rudifa:
I was intrigued by the difference between `.txtar` files intended for running with `testscript first.txtar` and those intended for running with `go test other.txtar`.

The files ran by `testscript` support the command `go` with a variety of arguments, as documented in `testscript -h`, in addition to commands, conditions and supporting files that are interpreted by the `small script engine in the testscript package`; files run by `go test` do not support the command `go`.

I suggest that this difference should be mentioned and documented in [testscript doc](https://pkg.go.dev/github.com/rogpeppe/go-internal@v1.11.0/testscript), for the benefit of newcomers.

mvdan:
The difference is https://pkg.go.dev/github.com/rogpeppe/go-internal/gotooltest. It's true that the gotooltest package is not linked from the testscript package's documentation, so I'd happily approve a small patch doing that.

mvdan:
Your PR above clarified this.

PR #249, opened 2024-03-10
testscript/doc: add `go` to the list of testscript commands

rudifa:
Fixes the [issue #234](https://github.com/rogpeppe/go-internal/issues/234).


mvdan:
Sorry, but this isn't right. What I tried to explain in https://github.com/rogpeppe/go-internal/issues/234#issuecomment-1985911411 is that using the gotooltest.Setup func is what adds the go command. It's not available in testscript by default.

mvdan (approved):
Thanks!

PR #251, opened 2024-05-11
testscript: Expose Environ() to custom commands

abhinav:
This adds an Environ method to TestScript
that looks and acts similarly to os.Environ.

This gives further visibility into the test script state
to custom command implementations,
allowing them to implement functionality that is otherwise limited
only to the built-in commands.

For example, combining this and MkAbs, a custom command can use os/exec
to run a command with the same environment as the test script,
with different behavior than the built-in TestScript.Exec method.


abhinav:
Gentle bump
@rogpeppe, @mvdan

PR #252, opened 2024-05-11
all: Add Go 1.22, drop Go 1.20

abhinav:
Switches CI to run against Go 1.21 and 1.22,
and bumps the minimum required Go version to 1.21.

Dropping 1.20 isn't strictly necessary;
this is just matching prior upgrades.
It can be added back if the maintainers prefer.


mvdan (approved):
Thanks!

PR #258, opened 2024-06-10
testscript: add Config.Files

rogpeppe:
This makes it possible to pass an arbitrary set of testscript files to be run instead of just a directory, making it possible for the testscript command to pass its command line arguments directly.

In order to check that all the files are actually tested, we need to make the test harness implement independent subtest failure, and it's useful to see the name of the test too so that we can see the name disambiguation logic at work, which makes for changes to some of the other tests too.

Note that the name deduping logic is somewhat improved from similar logic in cmd/testscript, in that it is always guaranteed to produce unique names even in the presence of filenames that look like deduped names.

### DIFF
--- a/testscript/testdata/testscript_explicit_files.txt
+++ b/testscript/testdata/testscript_explicit_files.txt
@@ -1,0 +1,26 @@
+# Check that we can pass an explicit set of files to be tested.
+! testscript -files foo.txtar x/bar.txtar y/bar.txtar 'y/bar#1.txtar'
+cmpenv stdout expect-stdout
+-- expect-stdout --
+** RUN foo **
+PASS
+** RUN bar **
+PASS
+** RUN bar#1 **
+> echoandexit 1 '' 'bar#1 failure'
+[stderr]
+bar#1 failure
+FAIL: $$WORK${/}y${/}bar.txtar:1: told to exit with code 1
+** RUN bar#1#1 **
+> echoandexit 1 '' 'bar#1#1 failure'
+[stderr]
+bar#1#1 failure
+FAIL: $$WORK${/}y${/}bar#1.txtar:1: told to exit with code 1
+-- foo.txtar --
+echoandexit 0 '' 'foo failure'
+-- x/bar.txtar --
+echoandexit 0 '' 'bar failure'
+-- y/bar.txtar --
+echoandexit 1 '' 'bar#1 failure'
+-- y/bar#1.txtar --
+echoandexit 1 '' 'bar#1#1 failure'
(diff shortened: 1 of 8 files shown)


testscript/testscript.go
```diff
@@ -307,10 +318,30 @@ func RunT(t T, p Params) {
...
+		// Take care to handle the situation where a name with a counter-like
+		// suffix already exists, for example:
+		//	a/foo.txt
+		//	b/foo.txtar
+		//	c/foo#1.txt
+		prefix := name
+		for i := 1; ; i++ {
+			if !names[name] {
+				break
+			}
```
abhinav, line 341: optional: this can be collapsed.
```suggestion
		for i := 1; names[name]; i++ {
```

abhinav (approved):
<sub>(Don't mind me. Just a drive-by review.)</sub>

Excellent! I've been meaning to open an issue to request something like.

testscript/testdata/testscript_explicit_files.txt
```diff
@@ -0,0 +1,26 @@
...
+** RUN bar#1 **
+> echoandexit 1 '' 'bar#1 failure'
+[stderr]
+bar#1 failure
+FAIL: $WORK/y/bar.txtar:1: told to exit with code 1
+** RUN bar#1#1 **
+> echoandexit 1 '' 'bar#1#1 failure'
+[stderr]
+bar#1#1 failure
+FAIL: $WORK/y/bar#1.txtar:1: told to exit with code 1
```
abhinav, line 18:
```suggestion
FAIL: $WORK${/}y${/}bar.txtar:1: told to exit with code 1
** RUN bar#1#1 **
> echoandexit 1 '' 'bar#1#1 failure'
[stderr]
bar#1#1 failure
FAIL: $WORK${/}y${/}bar#1.txtar:1: told to exit with code 1
```

testscript/testscript.go
rogpeppe: done

testscript/testdata/testscript_explicit_files.txt
rogpeppe: close, but needs `cmpenv`. done, thanks!

testscript/testscript.go
```diff
@@ -136,6 +137,11 @@ type Params struct {
 	// Dir is interpreted relative to the current test directory.
 	Dir string
 
+	// Files holds a set of script files. If Dir is empty and this
```
mvdan, line 140: say filenames to be clearer?

mvdan (approved):

testscript/testscript.go
rogpeppe: done

PR #259, opened 2024-06-10
cmd/testscript: do not create an extra temporary directory

rogpeppe:
See commit message.


mvdan (approved):
SGTM, although I did not review carefully.

PR #266, opened 2024-07-26
Bump Go dependencies

lucacome:
Bumps all go dependencies

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,9 @@
 module github.com/rogpeppe/go-internal
 
-go 1.22
+go 1.22.0
 
 require (
-	golang.org/x/mod v0.18.0
-	golang.org/x/sys v0.21.0
-	golang.org/x/tools v0.22.0
+	golang.org/x/mod v0.21.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/rogpeppe/go-internal
 
-go 1.21
+go 1.21.0
```
ldemailly, line 3: no no no :) keep it a "language minimal version" and not an "overly specific toolchain version" please
(just my 2 cents/crusade to keep/revert the go line to what it was before 1.21)

lucacome:
@mvdan @rogpeppe can this be merged?

go.mod
ldemailly: resolved? without an answer?

mvdan (approved):
Thanks - apologies for the slowness.

go.mod
mvdan: This really doesn't matter. Both mean almost exactly the same. If one dependency bumps us from 1.21 to 1.21.0, it's completely fine.

PR #267, opened 2024-07-26
Add dependabot

lucacome:
Adds dependabot to keep all the dependencies up to date

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+
+  - package-ecosystem: gomod
```
ldemailly, line 8: might want to group golang.org/x together so it's a single PR instead of 3 when these get bumped

mvdan:
Thanks but this is not really needed, given how updating all dependencies is literally one command :) At least in other Go projects I help maintain, the dependabot Go PRs tend to be more noise and annoyance than anything else. As a library, we also don't want to be updating dependencies immediately when new versions are out - we should only update as needed.

PR #268, opened 2024-08-08
testscript: add waitmatch command

mvdan:
(see commit message)

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -8,6 +8,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


testscript/cmd.go
```diff
@@ -652,6 +661,69 @@ func (ts *TestScript) waitBackground(checkStatus bool) {
 	ts.background = nil
 }
 
+// cmdWaitMatch waits until a background command prints a line to standard output
+// which matches the given regular expression. Once a match is found, the given
+// environment variable names are set to the subexpressions of the match.
+func (ts *TestScript) cmdWaitMatch(neg bool, args []string) {
+	if len(args) < 1 {
```
rogpeppe, line 668:
```suggestion
	if len(args) < 2 {
```

testscript/testscript.go
```diff
@@ -1027,22 +1044,26 @@ func (ts *TestScript) exec(command string, args ...string) (stdout, stderr strin
...
-		return nil, err
+		return nil, nil, err
 	}
 	cmd.Dir = ts.cd
 	cmd.Env = append(ts.env, "PWD="+ts.cd)
-	var stdoutBuf, stderrBuf strings.Builder
+	var stderrBuf strings.Builder
 	cmd.Stdin = strings.NewReader(ts.stdin)
-	cmd.Stdout = &stdoutBuf
+	stdoutr, stdoutw, err := os.Pipe()
```
rogpeppe, line 1059: Also, I'm not sure that os.Pipe is good enough here. If the process is producing a lot of output and the script never calls `waitmatch`, the pipe buffer could fill up and then the process will be blocked and potentially non-functioning as a result. I suspect we need to write everything to a buffer immediately, and then have a way to layer a reader on top of that.

Also, I wonder if we should make it possible to wait on stderr matches too? It's common for logged output to be printed to stderr and probably not uncommon to need to wait on that. Maybe waitmatch should wait on the combined output, although that might be tricky to arrange because we also need it to be split.

testscript/cmd.go
```diff
@@ -652,6 +661,69 @@ func (ts *TestScript) waitBackground(checkStatus bool) {
...
+			ts.Check(err)
+		}
+		if m != nil {
+			break
+		}
+	}
+}
+
+// readLine consumes enough bytes to read a line.
+func readLine(r io.Reader) ([]byte, error) {
```
rogpeppe, line 709: I don't get why we need to use this rather than (say) bufio.Scanner.

rogpeppe (changes requested):
Thanks for doing this! I'm not quite sure that the simple os.Pipe approach quite works though. Also, it would be good to add the command docs to `doc.go`.

PR #272, opened 2024-09-19
add goproxytest testing API, forward dirhash

mvdan:


goproxytest/proxy.go
```diff
@@ -45,26 +46,43 @@ type Server struct {
...
 	dir          string
+	logf         func(string, ...any)
 	modList      []module.Version
 	zipCache     par.Cache
 	archiveCache par.Cache
 }
 
-// StartProxy starts the Go module proxy listening on the given
+// NewTestServer is a wrapper around [NewServer] for use in Go tests.
+// Failure to start the server stops the test as failed,
```
rogpeppe, line 56: I wonder if it might be nice to document exactly which TB methods are called, making it easier for a caller that wishes to provide their own implementation to do so by implementing exactly those methods.

Although tbh I wouldn't mind exporting `newServer` as `NewServerWithLogger` (for example) which would make that easier again. I'm not sure that's really too complex of an API.

rogpeppe (approved):
LGTM with one thought, thanks!

goproxytest/proxy.go
mvdan: I thought about exporting newServer but it feels premature to do so. Practically every downstream use case has a testing.TB, and those who don't are main programs, where the existing logging is OK. I would rather do that when the need arises, or when we think more options would be useful.

Documented the TB methods used.

PR #281, opened 2024-11-24
testscript: phase out `func() int` in RunMain

mvdan:
(see commit message)

### DIFF
--- a/cmd/testscript/main.go
+++ b/cmd/testscript/main.go
@@ -41,37 +41,28 @@ func (e *envVarsFlag) Set(v string) error {
 }
 
 func main() {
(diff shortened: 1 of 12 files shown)


mvdan:
This has been bugging me for some time so I did something about it. I'm not particularly set on the name `Main` if you have a better suggestion. `Run` is taken, and I thought `TestMain` would be too confusing with the func that the user already has to write at the top level.

rogpeppe (approved):
LGTM but we should have at least one test that still exercises the old functionality, I think.

mvdan:
All addressed, thanks!

PR #290, opened 2025-02-24
testscript: remove temp dirs when finishing once again

mvdan:
(see commit message)

### DIFF
--- a/testscript/exe.go
+++ b/testscript/exe.go
@@ -50,50 +50,54 @@ func Main(m TestingM, commands map[string]func()) {
 	if mainf == nil {
 		// Unknown command; this is just the top-level execution of the
 		// test binary by "go test".
-
-		// Set up all commands in a directory, added in $PATH.
-		tmpdir, err := os.MkdirTemp("", "testscript-main")
-		if err != nil {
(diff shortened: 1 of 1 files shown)


testscript/exe.go
```diff
@@ -88,7 +89,11 @@ func Main(m TestingM, commands map[string]func()) {
 				ts.cmdExec(neg, append([]string{name}, args...))
 			}
 		}
-		os.Exit(m.Run())
+		exit := m.Run()
+		// The [os.Exit] below prevents defers from running, so we must clean up here.
+		// The defer above is still useful in case we fail before we get to this point.
+		cleanup()
```
rogpeppe, line 95: Alternatively (and arguably slightly cleaner) you could factor all the `Main` code into a separate function (e.g. `main1`) that returns int, and change `Main` to just do `os.Exit(main1(...))`

rogpeppe (approved):
LGTM with one small suggestion.

testscript/exe.go
mvdan: done; PTAL again as I ended up only factoring out part of the `Main` code, given that the other branch has no control over whether the user's command calls `os.Exit`.

rogpeppe (approved):

PR #295, opened 2025-04-09
testscript: protect against no-op uses of cmp

mvdan:
(see commit message)

### DIFF
--- a/testscript/cmd.go
+++ b/testscript/cmd.go
@@ -111,6 +111,11 @@ func (ts *TestScript) cmdCmpenv(neg bool, args []string) {
 
 func (ts *TestScript) doCmdCmp(neg bool, args []string, env bool) {
 	name1, name2 := args[0], args[1]
+	if name1 == name2 {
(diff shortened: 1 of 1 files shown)


testscript/cmd.go
```diff
@@ -111,6 +111,11 @@ func (ts *TestScript) cmdCmpenv(neg bool, args []string) {
 
 func (ts *TestScript) doCmdCmp(neg bool, args []string, env bool) {
 	name1, name2 := args[0], args[1]
+	if name1 == name2 {
+		// It's surprisingly easy to do e.g. `cmp foo.go foo.go` rather than
+		// `cmp foo.go foo.go.want`; protect against such obvious cases.
+		ts.Fatalf("cmp: comparing the same file twice")
```
rogpeppe, line 117: It's really only being compared once. How about "cannot compare a file against itself" ?

rogpeppe (approved):

testscript/cmd.go
mvdan: ah, I meant "twice" here as in "shows up twice in the argument list", but I agree it's ambiguous. done.

PR #302, opened 2026-03-17
update Go, fix, add /@latest to goproxytest

mvdan:


goproxytest/proxy.go
```diff
@@ -148,6 +148,41 @@ func (srv *Server) handler(w http.ResponseWriter, r *http.Request) {
...
+	if enc, ok := strings.CutSuffix(path, "/@latest"); ok {
+		modPath, err := module.UnescapePath(enc)
+		if err != nil {
+			srv.logf("go proxy_test: %v\n", err)
+			http.NotFound(w, r)
+			return
+		}
+		var best string
+		for _, m := range srv.modList {
+			if m.Path == modPath && !isPseudoVersion(m.Version) {
```
rogpeppe: I don't think this is quite right - according to https://go.dev/ref/mod#version-queries we ignore _all_ pre-release versions if there's release version, not just pseudo-versions. 

> The string `latest`, which selects the highest available release version. If there are no release versions, latest selects the highest pre-release version. If there are no tagged versions, `latest` selects a pseudo-version for the commit at the tip of the repository’s default branch.

https://github.com/cue-lang/cue/blob/97ee8d4c1b55d171a2077f26b58549014c995d05/internal/mod/modload/query.go#L120 implements this algorithm AIUI.

rogpeppe (commented):
Mostly LGTM but I think the version selection criteria aren't quite right.

goproxytest/proxy.go
mvdan: thanks for spotting - done

goproxytest/proxy.go
```diff
@@ -148,6 +148,50 @@ func (srv *Server) handler(w http.ResponseWriter, r *http.Request) {
...
 	}
 	path := strings.TrimPrefix(r.URL.Path, "/mod/")
+	if enc, ok := strings.CutSuffix(path, "/@latest"); ok {
+		modPath, err := module.UnescapePath(enc)
+		if err != nil {
+			srv.logf("go proxy_test: %v\n", err)
+			http.NotFound(w, r)
+			return
+		}
+		var bestRelease, bestAny string
```
rogpeppe: Personally, rather than inlining this logic into an already very long method, I'd probably do it in a standalone function where it's easy to see the logic and rationale in isolation. You could even write it in terms of `iter.Seq[string]` if you wanted to avoid creating an intermediate slice.

goproxytest/proxy_test.go
```diff
@@ -77,3 +80,29 @@ func TestSetup(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	var info struct {
+		Version string
+	}
+	if err := json.Unmarshal(body, &info); err != nil {
+		t.Fatalf("invalid JSON response: %v", err)
+	}
+	if info.Version != "v1.1.0" {
```
rogpeppe: There don't appear to be any prerelease versions of `fruit.com`; it would be nice to gain some confidence works with modules that have only prerelease versions and modules that have both, by adding some more test data and maybe having a two or three tests here.

rogpeppe (approved):
LGTM modulo a couple of suggestions below.

goproxytest/proxy_test.go
mvdan: done

goproxytest/proxy.go
mvdan: done; not sure if I got it exactly the way you envisioned, but we can always polish it up afterwards

PR #303, opened 2026-04-15
goproxytest: add Setup

rogpeppe:
The testscript command supports putting Go proxy files inside the same txtar file as everything else, but that functionality isn't easily available to regular testscript users. Address that by adding a `Setup` function that works in the same fashion as `gotooltest.Setup`, and make the testscript command use it.


mvdan (approved):
I'm not even sure that we need more tests for this, given that it's presumably already indirectly tested via cmd/testscript - up to you

PR #304, opened 2026-04-16
goproxytest: add overlay capability

rogpeppe:
It's useful to have a hybrid mode where some modules are served from the txtar contents but others come from the regular Go sources. We make goproxytest
capable of doing that: adding a file named `OVERLAY` to the top level `.gomodproxy` directory triggers that behavior.

### DIFF
--- a/goproxytest/setup.go
+++ b/goproxytest/setup.go
@@ -6,9 +6,13 @@ package goproxytest
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
+	"unicode"
 
+	"github.com/rogpeppe/go-internal/internal/goenv"
 	"github.com/rogpeppe/go-internal/testscript"
 )
 
@@ -24,6 +28,13 @@ const GoModProxyDir = ".gomodproxy"
 // It wraps p.Setup to start a proxy server when the directory is present,
 // set GOPROXY and GONOSUMDB appropriately, and shuts down
 // the server when the test completes.
+//
+// If a file exists in the top level of the [GoModProxyDir] directory
+// named "_enable_overlay", then instead of replacing the Go proxy entirely, the
+// existing Go proxy is used to serve modules that are not present in
+// the test script. In this case it uses the existing GOMODPROXY
+// download cache to avoid network downloads if it can, but takes care
+// not to pollute it with the localhost proxy server content.
 func Setup(p *testscript.Params) {
 	origSetup := p.Setup
 	p.Setup = func(env *testscript.Env) error {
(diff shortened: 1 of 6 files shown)


gotooltest/setup.go
```diff
@@ -74,18 +74,9 @@ func initGoEnv() (err error) {
...
-		"GOROOT",
-		"GOCACHE",
-		"GOPROXY",
-	)
-	if err != nil {
-		return fmt.Errorf("failed to determine environment from go command: %v\n%v", err, stderr)
-	}
-	if err := json.Unmarshal(eout.Bytes(), &goEnv); err != nil {
-		return fmt.Errorf("failed to unmarshal GOROOT and GOCACHE tags from go command out: %v\n%v", err, eout)
+	if err := goenv.Unmarshal(&goEnv); err != nil {
```
mvdan, line 77: FWIW this is a mild regression; you're now asking for all env vars, even though before we only fetched three. You could add `names ...string` to keep this in place, meaning that if the caller doesn't give any, all are fetched at extra cost.

goproxytest/testdata/setup/overlay.txt
```diff
@@ -0,0 +1,35 @@
...
+go get -d github.com/rogpeppe/go-internal/goproxytest
+
+-- go.mod --
+module mod
+
+-- .gomodproxy/OVERLAY --
+We want to be able to use regular Go
+modules as well as test-related ones.
+The contents of this file are ignored; only
+its presence is necessary.
```
mvdan, line 21: wrap this at 80-100 columns as usual, i.e. two lines. no need for four lines AFAICT

goproxytest/setup.go
```diff
@@ -24,6 +25,13 @@ const GoModProxyDir = ".gomodproxy"
 // It wraps p.Setup to start a proxy server when the directory is present,
 // set GOPROXY and GONOSUMDB appropriately, and shuts down
 // the server when the test completes.
+//
+// If a file exists in the top level of the [GoModProxyDir] directory
+// named "OVERLAY", then instead of replacing the Go proxy entirely, the
```
mvdan: Given that GoModProxyDir is `.gomodproxy`, I find the name `OVERLAY` a bit inconsistent and unnecessarily loud. Almost like it means to be an env var when it's not. I'd personally just go with `overlay`. It can't possibly be confused with a `modulepath_version` name.

goproxytest/setup.go
```diff
@@ -33,18 +41,46 @@ func Setup(p *testscript.Params) {
...
+				return err
+			}
+			if goEnv.GOMODCACHE == "" || goEnv.GOPROXY == "" {
+				// Shouldn't happen.
+				return fmt.Errorf("missing GOMODCACHE or GOPROXY from go env output")
 			}
-			env.Defer(srv.Close)
-			// Add GOPROXY after calling the original setup
-			// so that it overrides any GOPROXY set there.
+			// Look for local test modules first, falling back to the existing download cache.
```
mvdan:
```suggestion
			// Look for local test modules first, falling back to the existing module download cache.
```

for the sake of consistent naming with the Go docs

goproxytest/setup.go
```diff
@@ -33,18 +41,46 @@ func Setup(p *testscript.Params) {
...
-				"GOPROXY="+srv.URL,
-				"GONOSUMDB=*",
+				fmt.Sprintf("GOPROXY=%s,file://%s/cache/download,%s", srv.URL, goEnv.GOMODCACHE, goEnv.GOPROXY),
+				// Pass through the following variables on the grounds that
+				// they should be used when fetching external modules.
+				// Hopefully they will be immaterial because tests shouldn't
+				// be relying on modules that match them, but it seems safer
+				// to do this.
+				"GONOPROXY="+goEnv.GONOPROXY,
+				"GOPRIVATE="+goEnv.GOPRIVATE,
```
mvdan: forwarding `GOPRIVATE` is unnecessary; `go env` already shows how it propagates to the others:

```
$ GOPRIVATE=foobar go env | grep foobar
GONOPROXY='foobar'
GONOSUMDB='foobar'
GOPRIVATE='foobar'
```

We weren't passing it along before, and we don't need to pass it along for this, so I'd keep omitting it.

mvdan (approved):

gotooltest/setup.go
rogpeppe: Unfortunately that approach won't reduce the number of calls when two calls to goenv.Unmarshal both ask for distinct subsets of the variables. I've changed things so that we explicitly ask for all the env vars we might need - that's OK because this is an internal package and we know all the callers, and none of the variables are costly to calculate.

goproxytest/testdata/setup/overlay.txt
rogpeppe: Done

goproxytest/setup.go
rogpeppe: I've gone with `_enable_overlay` as discussed online

goproxytest/setup.go
rogpeppe: Done

goproxytest/setup.go
rogpeppe: Done

PR #305, opened 2026-04-17
goproxytest: make .mod and .info files optional

rogpeppe:
The `.mod` file is there because historical Go modules were not required to contain a go.mod file, but this is almost always redundant now, so just serve up the contents of the `go.mod` file if `.mod` is not present.

The `.info` file is similarly almost always redundant, so synthesize it when it's not present.

This should make providing synthetic Go module contents considerably more convenient.

Also, actually make the goproxytest log messages consistent this time.


goproxytest/proxy.go
```diff
@@ -386,6 +391,45 @@ func (srv *Server) readArchive(path, vers string) *txtar.Archive {
...
+}
+
+func archiveHasFile(a *txtar.Archive, name string) bool {
+	for _, f := range a.Files {
+		if f.Name == name {
+			return true
+		}
+	}
+	return false
+}
```
mvdan, line 422: join these two into a single func that returns `([]byte, bool)`

mvdan (approved):

goproxytest/proxy.go
rogpeppe: done

PR #306, opened 2026-06-09
testscript: add dynamic command capability

rogpeppe:
This allows a command to be able to add more commands, which is useful for the up-coming plugin functionality.

### DIFF
--- a/testscript/testdata/setcmd.txt
+++ b/testscript/testdata/setcmd.txt
@@ -1,0 +1,19 @@
+# A command registered with SetCmd becomes available for the rest of the test.
+register greet 'hello world'
+greet
+stdout '^hello world$'
+
+# A second registration works too, and re-registering replaces the command.
+register other 'goodbye'
+other
+stdout '^goodbye$'
+
+register greet 'hi again'
+greet
+stdout '^hi again$'
+
+# Registry a random name and check that it's not exec-able
+# (we're registering commands not executables)
+env name=d8e3ec30-d063-4f54-a9e6-b40eb830252a
+register $name x
+! exec $name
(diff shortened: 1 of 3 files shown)


testscript/testscript_test.go
```diff
@@ -175,8 +175,19 @@ func TestScripts(t *testing.T) {
 		Cmds: map[string]func(ts *TestScript, neg bool, args []string){
 			"setSpecialVal":    setSpecialVal,
 			"ensureSpecialVal": ensureSpecialVal,
-			"interrupt":        interrupt,
-			"waitfile":         waitFile,
+			"register": func(ts *TestScript, neg bool, args []string) {
+				// register <name> <text> dynamically defines a command
+				// that writes <text> to stdout, for the rest of the test.
+				if len(args) != 2 {
```
mvdan, line 181: also fail if negated? `! register ...` doesn't seem to make sense

testscript/testdata/setcmd.txt
```diff
@@ -0,0 +1,13 @@
...
+stdout '^hello world$'
+
+# A second registration works too, and re-registering replaces the command.
+register other 'goodbye'
+other
+stdout '^goodbye$'
+
+register greet 'hi again'
+greet
+stdout '^hi again$'
```
mvdan, line 13: nit: perhaps verify that if you do `register most_definitely_not_in_path 'whatever'`, then `exec most_definitely_not_in_path` fails

mvdan (approved):

testscript/testscript_test.go
rogpeppe: This is only for tests; I don't think it matters that much tbh.

rogpeppe: (note: we already ignore `neg` for all the other commands in this set - I don't think this is any different)

testscript/testdata/setcmd.txt
rogpeppe: done

PR #307, opened 2026-06-26
testscript/plugin: new package

rogpeppe:
This adds the capability for a testscript script to use a "plugin", an external binary that implements support for additional functionality not easily implemented with a regular command. Except when using the testscript command, it must be explicitly enabled by using `plugin.Setup`.

A plugin binary is run as an RPC server communicating over stdin/stdout, and can provide extra commands and environment variables to a testscript test. A classic use case is to run a scratch server, set an environment variable referring to its address, and tear it down when a test is done.

For now we use a custom RPC implementation so we can avoid polluting the Go API (unfortunately `net/rpc` requires that the types used are made public).

When `plugin` is invoked inside a testscript, e.g.

```
plugin foo
```

a plugin binary is searched for with the name `testscript-plugin-foo`.

For reference, we implement the existing Go proxy support as a plugin. In the future we may decide to fully factor out this support from the core testscript executable, although we'll have to be careful about backward compatibility concerns.

### DIFF
--- a/testscript/plugin/client.go
+++ b/testscript/plugin/client.go
@@ -1,0 +1,162 @@
+package plugin
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"sync"
+)
+
+// NewClient returns an implementation of [Interface] that issues JSON-RPC
+// requests to, and reads responses from, c. It performs an initial handshake
+// to fetch the plugin's [PluginInfo]. The connection c is closed when the
+// returned value's Close method is called.
+//
+// The returned value is safe for concurrent use; in particular the test
+// instances it creates may be used concurrently.
+func NewClient(c io.ReadWriteCloser) (Interface, error) {
+	cl := &client{
+		c:       c,
+		enc:     json.NewEncoder(c),
+		pending: make(map[uint64]chan rpcResponse),
+	}
+	go cl.readLoop()
+	var res infoResult
+	if err := cl.call(methodInfo, infoArgs{Version: protocolVersion}, &res); err != nil {
+		c.Close()
+		return nil, fmt.Errorf("plugin handshake failed: %v", err)
+	}
+	if res.Version > protocolVersion {
+		c.Close()
+		return nil, fmt.Errorf("plugin selected protocol version %d, greater than supported version %d", res.Version, protocolVersion)
+	}
+	cl.info = res.Info
+	return cl, nil
+}
+
+type client struct {
+	c    io.ReadWriteCloser
+	info PluginInfo
+
+	// mu guards the fields below it.
+	mu      sync.Mutex
+	enc     *json.Encoder
+	seq     uint64
+	pending map[uint64]chan rpcResponse
+	err     error // set once the read loop terminates
+}
+
+// readLoop reads responses from the connection and delivers each to the
+// goroutine waiting on its request id. When the connection fails it records
+// the error and wakes all waiting callers.
+func (cl *client) readLoop() {
+	dec := json.NewDecoder(cl.c)
+	for {
+		var resp rpcResponse
+		if err := dec.Decode(&resp); err != nil {
+			cl.mu.Lock()
+			cl.err = err
+			pending := cl.pending
+			cl.pending = nil
+			cl.mu.Unlock()
+			for _, ch := range pending {
+				close(ch)
+			}
+			return
+		}
+		cl.mu.Lock()
+		ch := cl.pending[resp.ID]
+		delete(cl.pending, resp.ID)
+		cl.mu.Unlock()
+		if ch != nil {
+			ch <- resp
+		}
+	}
+}
+
+// call sends a request and waits for its response, unmarshaling the result
+// into result (if non-nil). A non-empty response error is returned as an error.
+// TODO we could pass a context in here to enable timing out.
+func (cl *client) call(method string, arg, result any) error {
+	var params json.RawMessage
+	if arg != nil {
+		data, err := json.Marshal(arg)
+		if err != nil {
+			return err
+		}
+		params = data
+	}
+	ch := make(chan rpcResponse, 1)
+	cl.mu.Lock()
+	if cl.err != nil {
+		err := cl.err
+		cl.mu.Unlock()
+		return err
+	}
+	cl.seq++
+	id := cl.seq
+	cl.pending[id] = ch
+	err := cl.enc.Encode(rpcRequest{ID: id, Method: method, Params: params})
+	if err != nil {
+		delete(cl.pending, id)
+		cl.mu.Unlock()
+		return err
+	}
+	cl.mu.Unlock()
+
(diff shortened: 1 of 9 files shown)


testscript/testscript.go
```diff
@@ -521,6 +521,14 @@ func (ts *TestScript) setup() string {
...
 	if runtime.GOOS == "windows" {
 		env.Vars = append(env.Vars, "exe=.exe")
+		// SystemRoot (and windir) are required for networking to work in
+		// subprocesses: without them Winsock cannot load its service
+		// providers and any socket operation fails.
+		for _, name := range []string{"SystemRoot", "windir"} {
+			if v := os.Getenv(name); v != "" {
+				env.Vars = append(env.Vars, name+"="+v)
+			}
+		}
```
mvdan: we removed SYSTEMROOT in 3030644a6651fc95902ebe27e31581838861b233, why are you re-adding these? note how os/exec always passes this one along now.

testscript/plugin/contract.go
```diff
@@ -0,0 +1,146 @@
...
+type CmdInfo struct {
+	// RequiredEnv holds the set of environment variable names that the
+	// command needs passed to it. Only these variables are sent to the
+	// plugin when the command is run.
+	RequiredEnv map[string]bool
+
+	// NeedStderr and NeedStdout indicate whether the command produces
+	// standard error or standard output respectively.
+	NeedStderr bool
+	NeedStdout bool
```
mvdan: these never seem to be read by the non-test logic? either wire them properly or remove

testscript/plugin/client.go
```diff
@@ -0,0 +1,157 @@
...
+		cl.mu.Unlock()
+		if ch != nil {
+			ch <- resp
+		}
+	}
+}
+
+// call sends a request and waits for its response, unmarshaling the result
+// into result (if non-nil). A non-empty response error is returned as an error.
+func (cl *client) call(method string, arg, result any) error {
```
mvdan, line 81: my instinct is that this should take a context to be able to use timeouts/deadlines or cancel, as otherwise a plugin over RPC can hang or take a long time to reply, hanging the caller too.

testscript/plugin/plugin.go
```diff
@@ -0,0 +1,226 @@
...
+	cmd := exec.Command(path)
+	w, err := cmd.StdinPipe()
+	if err != nil {
+		return nil, err
+	}
+	r, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+	cmd.Stderr = os.Stderr
```
mvdan: note that if you send stderr straight to os.Stderr, any failures happening during `go test` will be mixed rather than being matched to the correct test run via e.g. `testing.T.Log`. perhaps not a big deal.

testscript/plugin/contract.go
```diff
@@ -0,0 +1,146 @@
...
+type CmdParams struct {
+	// CmdName is the name of the command being run.
+	CmdName string
+
+	// Args holds the command's arguments (not including the command name).
+	Args []string
+
+	// CWD is the absolute path of the current working directory the
+	// command should run in.
+	CWD string
```
mvdan: I'd personally name this Dir, for consistency with os/exec and most other Go APIs

testscript/plugin/contract.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Close shuts the instance down, releasing any resources held for the
+	// test.
+	Close()
+}
+
+// CmdParams holds the parameters for running a plugin command.
+type CmdParams struct {
+	// CmdName is the name of the command being run.
+	CmdName string
```
mvdan: nit: just Name? the other fields are not named CmdArgs etc.

testscript/plugin/contract.go
```diff
@@ -0,0 +1,146 @@
...
+	// Args holds the command's arguments (not including the command name).
+	Args []string
+
+	// CWD is the absolute path of the current working directory the
+	// command should run in.
+	CWD string
+
+	// Environ holds the values of the variables named in the command's
+	// [CmdInfo.RequiredEnv], as resolved in the test environment.
+	Environ map[string]string
```
mvdan: the name Environ in Go makes me think of `os.Environ`, which is a `[]string`, not a map. so if you really do want to use a map, I'd use a different name like `Env`

testscript/plugin/contract.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Stdin holds the standard input for the command.
+	Stdin []byte
+}
```
mvdan, line 128: it looks like this plugin system doesn't support background commands. that's probably fine for the RPC model, but do we test that these are rejected properly?

testscript/plugin/contract.go
```diff
@@ -0,0 +1,146 @@
...
+	// NewTestInstance creates a new instance of the plugin for a single
+	// running test. Multiple instances may exist concurrently.
+	NewTestInstance(TestParams) (TestInstance, error)
+
+	// Close shuts the plugin down, releasing any global resources.
+	Close()
+}
+
+// PluginInfo holds static information about a plugin.
+type PluginInfo struct {
```
mvdan, line 48: personally I'd add something like a `Version int` here, starting at 1, such that we can later bump this if we need to do any significant changes.

mvdan (approved):

testscript/plugin/client.go
rogpeppe: yeah, it can block indefinitely but currently there's no notion of context or timeout in command interface, so I think it's OK to just let it block for now, as with any other command implementation. We could add timeouts as an enhancement later.

rogpeppe: I added a TODO

testscript/plugin/contract.go
rogpeppe: wired up and changed name to be more appropriate.

testscript/plugin/contract.go
rogpeppe: done, a bit differently, allowing for version negotiation.

testscript/plugin/contract.go
rogpeppe: I'm not sure there's much need: `&` is only supported for `exec` commands not commands in general.

testscript/plugin/plugin.go
rogpeppe: Done

testscript/plugin/contract.go
rogpeppe: done

testscript/plugin/contract.go
rogpeppe: done

testscript/plugin/contract.go
rogpeppe: Done

testscript/testscript.go
rogpeppe: done

PR #309, opened 2026-09-22
testscript: retry Windows command cleanup

### LINKED ISSUE #283: gotooltest: testing flake on Windows: "access is denied"
From https://github.com/rogpeppe/go-internal/actions/runs/12123750949/job/34054672393?pr=282

```
--- FAIL: TestSimple (0.80s)
    --- FAIL: TestSimple/cover (13.19s)
        testscript.go:584: # The module uses testscript itself.
            # Use the checked out module, based on where the test binary ran. (7.894s)
            # First, a 'go test' run without coverage. (2.677s)
            > go test -vet=off
            [stdout]
            PASS
            2024/12/06 21:03:20 cannot delete temporary directory: remove $WORK\.tmp\testscript-main7[56](https://github.com/rogpeppe/go-internal/actions/runs/12123750949/job/34054672393?pr=282#step:4:57)463866\bin\foo.exe: Access is denied.
            exit status 2
            FAIL	test	0.933s
            
            [exit status 1]
            FAIL: testdata\cover.txt:9: unexpected go command failure
            
FAIL
```

Discovered whilst testing https://github.com/rogpeppe/go-internal/pull/282 at commit 04a2a257aa6fc8e8bee48a527990a1064673251e

sb123sb123:
Fixes https://github.com/rogpeppe/go-internal/issues/283

## Problem

On Windows, testscript can fail after a command subprocess has completed because the temporary directory cleanup sees `Access is denied` for the copied command executable.

## Cause

`testingMRun` used `os.RemoveAll` directly. Windows can transiently deny removal while an executable launched from the temporary command directory is still being released.

## Fix

Use the repository's existing `robustio.RemoveAll`, which retries transient Windows filesystem errors. The regression test keeps a copied test command running while `testingMRun` performs cleanup and verifies that cleanup succeeds after the process releases the executable.

## Tests

- Before the fix, the focused Windows regression reproduced: `cannot delete temporary directory ... holdopen.exe: Access is denied`.
- After the fix: `go test ./testscript -run '^TestMainCleanupRetriesAccessDenied$' -count=1`.
- After the fix: `go test ./testscript -run '^(TestMainCleanupRetriesAccessDenied|TestScripts)$' -count=1`.
- `go vet ./testscript`.
- `gofmt` check.
- Full repository compile-only check: `go test -run '^$' ./...`.
- The existing `gotooltest` coverage scenario passes after the fix.

## Limitations

The remote Windows host could not download the repository's Go modules from `proxy.golang.org`. The full runtime suite therefore has environment-related failures in tests that require dependency downloads or assert the default `GOPROXY` value; exact task-owned dependency snapshots were used for the focused and compile-only validations. `golangci-lint` was not installed on the host.

## AI assistance

This pull request was prepared with assistance from OpenAI Codex. I reviewed the issue, implementation, regression test, and validation results.


testscript/exe_test.go
```diff
@@ -0,0 +1,78 @@
...
+	if err := os.Setenv(holdOpenReadyEnv, ready); err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() {
+		if hadOldReady {
+			_ = os.Setenv(holdOpenReadyEnv, oldReady)
+		} else {
+			_ = os.Unsetenv(holdOpenReadyEnv)
+		}
+	})
```
rogpeppe: ISTM we can just use `testing.T.Setenv` here:

```suggestion
    t.Setenv(holdOpenReadEnv, ready)
```

testscript/exe_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+const holdOpenReadyEnv = "TESTSCRIPT_HOLDOPEN_READY"
+
+type testM struct {
+	run func() int
+}
+
+func (m testM) Run() int {
+	return m.run()
+}
```
rogpeppe:
```suggestion
type testM func() int

func (m testM) Run() int {
	return m()
}
```

testscript/exe_test.go
```diff
@@ -0,0 +1,78 @@
...
+func holdOpen() {
+	if err := os.WriteFile(os.Getenv(holdOpenReadyEnv), nil, 0o666); err != nil {
+		os.Exit(2)
+	}
+	time.Sleep(750 * time.Millisecond)
+}
+
+func TestMainCleanupRetriesAccessDenied(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		t.Skip("Windows executable sharing semantics are required")
```
rogpeppe: I don't see why we only need to run this test under Windows. ISTM that it should pass OK under any OS.

testscript/exe_test.go
```diff
@@ -0,0 +1,78 @@
...
+			_ = os.Setenv(holdOpenReadyEnv, oldReady)
+		} else {
+			_ = os.Unsetenv(holdOpenReadyEnv)
+		}
+	})
+
+	var cmd *exec.Cmd
+	code := testingMRun(testM{run: func() int {
+		cmd = exec.Command("holdopen")
+		if err := cmd.Start(); err != nil {
```
rogpeppe:
```suggestion
		// Start a command but don't wait for it to finish.
		// This triggers the Windows behaviour of refusing
		// to let the directory be removed until the command
		// is complete.
		if err := cmd.Start(); err != nil {
```

testscript/exe_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (m testM) Run() int {
+	return m.run()
+}
+
+func holdOpen() {
+	if err := os.WriteFile(os.Getenv(holdOpenReadyEnv), nil, 0o666); err != nil {
+		os.Exit(2)
+	}
+	time.Sleep(750 * time.Millisecond)
```
rogpeppe: Given that the test will always wait for this amount of time under Windows, I think this seems like quite a long time. Maybe 100ms might be more appropriate. I don't think it matters that very occasionally we might not be testing the desired code path.

testscript/exe_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+type testM struct {
+	run func() int
+}
+
+func (m testM) Run() int {
+	return m.run()
+}
+
+func holdOpen() {
```
rogpeppe: Let's put this helper function towards the end of the file rather than before the things that call it.

testscript/exe_test.go
```diff
@@ -0,0 +1,78 @@
+// Copyright 2026 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testscript
```
rogpeppe: Ideally it would be good to add this test in a PR without the fix, with an assertion that it fails. Then add a PR that fixes it. That way we can be sure that the fix really does work and that the test really does test the correct thing. That's particularly true when we're targeting platforms that we're not actually developing on (I can't easily test that the fix does properly fix the issue because I don't run Windows).

rogpeppe (commented):
Thanks for contributing this and adding the test. LGTM in principle, but I'd like to be more sure the fix really is sufficient, hence my suggestion about adding the test in a separate PR (ideally we'd have one depending on the other, with CI passing on both, but I'm not sure that's possible in GitHub currently).

